PR #2, opened 2025-10-07
Multi go module setup

rytswd:
This introduces the multi Go module setup, where each module can have separate dependencies.

For local development, you can ensure LSP checks at one go.mod by opening at the module root directory, or use `go.work` and `go workspace` setup to allow opening up from the repository root. There is some mention about Go workspace usage included in the document as well.

~~NOTE: I have added test impl of `main.go`, as well as `go.work` which is only meant to be used locally. This is just to demonstrate how we can handle the configuration, and that commit should be removed before this can be merged.~~

This has been removed.

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -72,7 +72,7 @@ The operator is organized into three independent Go modules with clear separatio
 #### Cluster Handler Module (`pkg/cluster-handler`)
 - **Purpose**: High-level cluster orchestration
 - **Responsibilities**:
-  - Manages MultigresCluster custom resource
+  - Uses the `MultigresCluster` custom resource as its primary input
   - Coordinates creation of child resources (MultiGateway, MultiOrch, MultiPooler, Etcd, Cell)
   - Aggregates status from child resources
   - Handles cluster-wide lifecycle (finalizers, upgrades)
@@ -81,7 +81,7 @@ The operator is organized into three independent Go modules with clear separatio
 #### Data Handler Module (`pkg/data-handler`)
 - **Purpose**: Data plane management and topology
 - **Responsibilities**:
-  - Manages Cell custom resource (data topology abstraction)
+  - Uses the `Cell` custom resource as its primary input (data topology abstraction)
   - Handles data plane configuration and routing
   - Coordinates with Vitess/Multigres data components
 - **Independence**: Data plane logic isolated from control plane and resource management
@@ -89,7 +89,7 @@ The operator is organized into three independent Go modules with clear separatio
 #### Resource Handler Module (`pkg/resource-handler`)
 - **Purpose**: Individual component resource management
 - **Responsibilities**:
-  - Manages component CRDs: MultiGateway, MultiOrch, MultiPooler, Etcd
+  - Uses the `MultiGateway`, `MultiOrch`, `MultiPooler`, `Etcd` custom resources as its primary inputs
   - Creates and reconciles Kubernetes resources (Deployments, StatefulSets, Services, HPAs)
   - Implements pure resource builder functions
   - Component-specific health checking and status reporting


docs/implementation-guide.md
```diff
@@ -22,10 +22,14 @@
...
 
 ### Dependency Management
-- **Go Modules**: `go.mod` and `go.sum` for dependency tracking
-- **Version Pinning**: Pin controller-runtime and client-go to compatible versions
-- **Vendor Directory**: Not used - rely on module cache
-- **Updating Dependencies**: Use `go get -u` carefully, test thoroughly after updates
+- **Per-Module Dependencies**: Each module (`cluster-handler`, `data-handler`, `resource-handler`) has its own `go.mod` and `go.sum`
+- **Version Pinning**: Pin controller-runtime and client-go to compatible versions in each module
+- **Go Workspaces (Local Only)**:
+  - Create `go.work` locally for easier multi-module development: `go work init ./pkg/cluster-handler ./pkg/data-handler ./pkg/resource-handler`
```
rcambrj, line 28: Recommend a short PSA in the pull request description, something like:

> Dear fellow developers, after merging you should run:
> ```
> go work init ./pkg/cluster-handler ./pkg/data-handler ./pkg/resource-handler
> ```

...to permit cross-module references without publishing

docs/architecture.md
```diff
@@ -40,36 +40,68 @@
...
-- **Label Management**: Consistent label generation for resource selection
-- **No Side Effects**: Same input always produces same output
+### Multi-Module Architecture
+
+The operator is organized into three independent Go modules with clear separation of concerns. Developers can use **Go workspaces** locally (`go.work`) for easier cross-module development, though this file is not committed to the repository:
+
+#### Cluster Handler Module (`pkg/cluster-handler`)
+- **Purpose**: High-level cluster orchestration
+- **Responsibilities**:
+  - Manages MultigresCluster custom resource
```
rcambrj: I'm not sure that `manages` is the correct term here (and for the same responsibility in the `data-handler` section). I'd argue that this is the _input_ for this module and thus is managed by something else.

docs/implementation-guide.md
```diff
@@ -145,12 +151,13 @@ make manifests
 make install
 ```
 
-**Controller Changes** (`internal/controller/`):
-- Run `make test` frequently during development
+**Controller Changes** (`pkg/{module}/controller/`):
+- Run tests in the specific module: `cd pkg/resource-handler && go test ./...`
 - Integration tests catch reconciliation bugs early
 - Use `make run` for quick iteration (no docker build needed)
```
rcambrj, line 157: I realise that this line is unchanged - just pointing out that we should remember to create a Makefile with a `run` target :D

docs/implementation-guide.md
rytswd: Maybe we should have a make definition, or simple script for this -- but at the same time, if you are working on a single module (and you should be usually working for a single module only), I don't think this is necessary.

The main difficulty I can see is when you use IDEs like VSCode -- without go.work in place, it would fail to find the right go.mod to start the LSP with.

What do you think we can proceed with?

rcambrj: A Makefile target would be nice too, I've no objections against that.

This being a one time only task when setting up your working directory for development, any existing devs will need to perform this task once after merging this change - it's just a nice courtesy to make that obvious in the PR's description.

docs/architecture.md
rcambrj: As discussed, let's go with something a bit more descriptive. I propose:

> Uses the `MultigresCluster` custom resource as its primary input

and in the following sections...

> Uses the `Cell` custom resource as its primary input (data topology abstraction)

> Uses the `MultiGateway`, `MultiOrch`, `MultiPooler`, `Etcd` custom resources as its primary inputs

docs/implementation-guide.md
rytswd: Makefile is used throughout Kubernetes ecosystem, and we will likely end up with using it, so it would probably make sense to add an entry there.

As to the clarity for new developer onboarding, do you mean the PR template for all PRs created, or the description of this PR?

docs/architecture.md
rytswd:
```suggestion
  - Uses the `MultigresCluster` custom resource as its primary input
```

docs/architecture.md
```diff
@@ -40,36 +40,68 @@
...
+  - Coordinates creation of child resources (MultiGateway, MultiOrch, MultiPooler, Etcd, Cell)
+  - Aggregates status from child resources
+  - Handles cluster-wide lifecycle (finalizers, upgrades)
+- **Independence**: Can be tested and versioned separately from resource handlers
+
+#### Data Handler Module (`pkg/data-handler`)
+- **Purpose**: Data plane management and topology
+- **Responsibilities**:
+  - Manages Cell custom resource (data topology abstraction)
```
rytswd:
```suggestion
  - Uses the `Cell` custom resource as its primary input (data topology abstraction)
```

docs/architecture.md
```diff
@@ -40,36 +40,68 @@
...
+  - Handles data plane configuration and routing
+  - Coordinates with Vitess/Multigres data components
+- **Independence**: Data plane logic isolated from control plane and resource management
+
+#### Resource Handler Module (`pkg/resource-handler`)
+- **Purpose**: Individual component resource management
+- **Responsibilities**:
+  - Manages component CRDs: MultiGateway, MultiOrch, MultiPooler, Etcd
```
rytswd:
```suggestion
  - Uses the `MultiGateway`, `MultiOrch`, `MultiPooler`, `Etcd` custom resources as its primary inputs
```

docs/implementation-guide.md
rcambrj: Just the description of this PR :)

rytswd: Thanks, I've updated the description with the following:

> For local development, you can ensure LSP checks at one go.mod by opening at the module root directory, or use `go.work` and `go workspace` setup to allow opening up from the repository root. There is some mention about Go workspace usage included in the document as well.

PR #3, opened 2025-10-07
initial structure and CI/CD

rcambrj:
* brings CI/CD pipeline over from temporary repository
* adjusted go module location to root
* simplified build targets (just one hardcoded now)
* added build-time variables providing `version` and `commit`

here's an [example release](https://github.com/numtide/multigres-operator/releases/tag/test)
here's an [example container image](https://github.com/numtide/multigres-operator/pkgs/container/multigres-operator)

### DIFF
--- a/.github/workflows/build-and-release.yaml
+++ b/.github/workflows/build-and-release.yaml
@@ -1,0 +1,136 @@
+name: Build & Release
+
+on:
+  workflow_call:
+    inputs:
+      architectures:
+        description: A JSON list of GOARCH values for which to build
+        default: '["amd64", "arm64"]'
+        required: false
+        type: string
+      push-container-image:
+        description: Whether to push the resulting container image to the registry
+        default: false
+        required: false
+        type: boolean
+      create-release:
+        description: Whether to create a Github release and attach the artifacts to it
+        default: false
+        required: false
+        type: boolean
+
+permissions:
+  contents: write
+  packages: write
+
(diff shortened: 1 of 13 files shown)


.github/workflows/build-and-release.yaml
rytswd, line 1: Can you make this reusable workflow defined somewhere other than `.github/workflows` directory? Although the details here are very relevant, it would help discoverability of what GitHub Action jobs we've got in the repo by moving all the reusable ones out of this directory

rytswd (approved):
LGTM, just one minor suggestion

.github/workflows/build-and-release.yaml
rcambrj: I had tried originally putting the reusable workflow in `.github/foo` (the exact name isn't important) and Github did not like that. For confirmation, I just asked GPT-5:

> given a repository on github with workflows, can a `on.workflow_call` reusable workflow be located anywhere other than `.github/workflows`?
>
> No. GitHub Actions only recognizes workflow files located in the `.github/workflows` directory of the target repository (or the organization’s .github repository). Reusable workflows triggered with on: workflow_call must live there.

Another strategy would be to use "composite actions", which can apparently live anywhere in the repository. Would you prefer that?

rytswd: Okay if that's the case it's fine as is. That's a strange limitation!

PR #6, opened 2025-10-08
implement basic automatic code scanning

rcambrj:
* scans dependencies with `actions/dependency-review-action` (only during pull request workflow, and only scans new dependencies) - [this is what it looks like](https://github.com/numtide/multigres-operator/actions/runs/18344810858/job/52248714761)

  Screenshot 2025-10-08 at 14 35 58

* scans all code (including Go, but I suspect excluding nix) with `github/codeql-action` - [this is what it looks like](https://github.com/numtide/multigres-operator/pull/4#pullrequestreview-3314623235)

  Screenshot 2025-10-08 at 14 52 59

* scans the built container image with `anchore/scan-action` (a wrapper around `grype`) - [this is what it looks like](https://github.com/numtide/multigres-operator/security/code-scanning?query=is%3Aopen+branch%3Amain+pr%3A4+)

  Screenshot 2025-10-08 at 15 07 38

* and locks all `uses:` to a fixed commit hash whilst retaining a human-friendly reference to the version in a comment

### Bits I don't like (could be improved later)

* there is no nix code scanning - [`vulnix`](https://github.com/nix-community/vulnix) exists there is no github action that I could find, and given that nix plays only a devshell role, I figured we could tackle this another day.
* the `grype` container image scanner requires that the container image be uploaded (pushed) and then downloaded (pulled) before scanning. For this reason, the container image gets pushed even when the workflow's input `push-container-image` is `false` - only the `latest` and tag versions get pushed when `push-container-image` is true.

### DIFF
--- a/.github/workflows/build-and-release.yaml
+++ b/.github/workflows/build-and-release.yaml
@@ -22,24 +22,67 @@ on:
 permissions:
   contents: write
   packages: write
+  security-events: write
 
 jobs:
+  scan-dependencies:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out code
+        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+
+      - uses: actions/dependency-review-action@56339e523c0409420f6c2c9a2f4292bbb3c07dd3 # v4.8.0
+        # this action requires a base and head
+        # ensure that trunk changes require pull request checks to pass
+        if: ${{ github.event_name == 'pull_request' }}
+
+  scan-codeql:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out code
+        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+
+      - uses: github/codeql-action/init@e296a935590eb16afc0c0108289f68c87e2a89a5 # v4.30.7
+        with:
+          languages: actions,go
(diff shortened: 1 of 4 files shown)


.github/workflows/build-and-release.yaml
```diff
@@ -22,24 +22,67 @@ on:
...
+      - uses: actions/dependency-review-action@56339e523c0409420f6c2c9a2f4292bbb3c07dd3 # v4.8.0
+        # this action requires a base and head
+        # ensure that trunk changes require pull request checks to pass
+        if: ${{ github.event_name == 'pull_request' }}
+
+  scan-codeql:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out code
+        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
```
rytswd, line 43: I think it is common to use a version tag than commit sha with GitHub Actions, but I'm fine with this. We'd need a quick script or some logic to update this though

rcambrj: I came across this approach while researching these code scanning tools, where I came saw code written by more security-conscious folk. The concern with tags, as usual, is that they can change. You would normally have a lockfile with a hash for dependencies, but in the case of github workflows there is none. This is the next best thing.

I accept that it's a bit janky and appreciate that this requires some flexibility from developers' POV, and I agree that an updater tool would be great.

PR #8, opened 2025-10-09
Add planning doc for Cell CRD

rytswd:
Cell CRD handling will require a dedicated module, and this will be one of the confusing part.

The main target is to get the CRD in place first, and relevant resource setup in place.

### DIFF
--- a/plans/phase-1/cell-management-operator.md
+++ b/plans/phase-1/cell-management-operator.md
@@ -125,7 +125,7 @@ func (r *CellReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
 
 ## Error Handling
 
-- **Etcd unreachable**: Set `TopoServerReachable: false`, requeue with backoff
+- **Etcd unreachable**: Set `spec.topoServerReachable: false`, requeue with backoff
 - **Invalid component references**: Log warning, continue (components may not exist yet)
 - **Registration failure**: Update Condition with error details, requeue
 


plans/phase-1/cell-management-operator.md
```diff
@@ -0,0 +1,226 @@
+---
+title: Create Cell management Go module as Multigres aware operator internal
```
rcambrj: Does this line make sense? What does `internal` mean at the end?

plans/phase-1/cell-management-operator.md
```diff
@@ -0,0 +1,226 @@
...
+
+## Finalizer Handling
+
+- Add finalizer `cell.multigres.com/finalizer` on Cell creation
+- On deletion, remove Cell entry from Topo Server before removing finalizer
+- Ensures Cell is properly deregistered before CR is deleted
+
+## Error Handling
+
+- **Etcd unreachable**: Set `TopoServerReachable: false`, requeue with backoff
```
rcambrj: is this a custom resource status attribute? (please make it clear what kind of attribute this is)

plans/phase-1/cell-management-operator.md
rytswd: Sorry and thanks for the catch, it got mixed up with something I had a draft for 🙏

```suggestion
title: Create Cell management Go module with upstream Multigres dependency
```

plans/phase-1/cell-management-operator.md
rytswd: This would be just a field inside the Cell CRD. The status stanza should not be used for operator logic, it's purely for display

rcambrj: if the list of cells is sent each time, this might not be relevant. A more relevant field would be at a topo-level and would reflect the last time at which the list was successfully sent.

rytswd: Unless we have external validation logic using webhook, there could be some conflict between resources, and having a field like this would be helpful to identify the cluster state.
There are probably better ways to do this though -- it's only meant to requeue early

rcambrj: Understood. Back to my original request: would it be agreeable to declare where this `TopoServerReachable` would be set?

rytswd: I don't quite get what you mean -- it won't be a part of status as this is used for controller logic, and thus will be in spec. Unless otherwise mentioned it would be top level I'd think -- or do you think there is a better place?

rcambrj: Right! That's what I mean :)

```suggestion
- **Etcd unreachable**: Set `spec.TopoServerReachable: false`, requeue with backoff
```

rytswd: I updated this with 2648080

rcambrj: ship it!

PR #9, opened 2025-10-09
Add planning doc for defining CRDs

rytswd:
Add the planning doc for CRD spec handling using kubebuilder

### DIFF
--- a/plans/phase-1/create-child-resource-crds.md
+++ b/plans/phase-1/create-child-resource-crds.md
@@ -116,7 +116,6 @@ Each component spec should include:
 
 **Example Affinity Use Cases**:
 - **Anti-affinity for HA**: Spread etcd replicas across different nodes/zones to survive node failures
-- **Co-location**: Place MultiPooler near Postgres instances for low latency
 - **Node isolation**: Dedicate specific nodes for database workloads using node affinity
 - **Zone spreading**: Distribute replicas across availability zones using topology spread constraints
 


plans/phase-1/create-child-resource-crds.md
```diff
@@ -0,0 +1,394 @@
...
+- kubebuilder
+---
+
+# Summary
+
+Create individual CRDs for each Multigres component (`MultiGateway`, `MultiOrch`, `MultiPooler`, `Etcd`) and a parent `MultigresCluster` CRD. Each component CRD can be deployed independently or as part of a `MultigresCluster`. This provides flexibility for users to manage components individually while still supporting convenient cluster-level management.
+
+This document focuses on defining the API types and CRD structure. Actual controller implementation and `MultigresCluster` parent-child creation logic are covered in separate documents.
+
+# Motivation
```
rcambrj, line 19: This section appears to skip one of the advantages we discussed, which is to separate the concerns of the `data-handler` and the `resource-handler`. I should add a bullet point about that.

plans/phase-1/create-child-resource-crds.md
```diff
@@ -0,0 +1,394 @@
...
+- **Affinity**: Optional pod affinity, anti-affinity, and node affinity rules
+- **Tolerations**: Optional tolerations for node taints
+- **NodeSelector**: Optional node selector labels
+- **TopologySpreadConstraints**: Optional constraints for pod distribution across topology domains
+- **Storage**: (For StatefulSet-based components like Etcd, MultiPooler)
+- **Component-specific fields**: Any unique configuration needs
+
+**Example Affinity Use Cases**:
+- **Anti-affinity for HA**: Spread etcd replicas across different nodes/zones to survive node failures
+- **Co-location**: Place MultiPooler near Postgres instances for low latency
```
rcambrj: MultiPooler and postgres have a 1-1 relationship. Might they be run in a single pod?

plans/phase-1/create-child-resource-crds.md
rytswd: This was intentional, because the use of separate Go module is only necessary if we were to generate a Cell CR. This should give the clarity of https://github.com/numtide/multigres-operator/pull/11 needed, but not so much on the actual implementation details

plans/phase-1/create-child-resource-crds.md
rytswd: Thanks for the catch, affinity is only for zone and node handling - fixed with 8cdb2c2

plans/phase-1/create-child-resource-crds.md
rcambrj: I don't fully understand the motivation to leave this point out of the motivation, but happy that you've considered it!

rytswd: The fact that there are multiple CRDs has nothing to do with the module split up,. Vitess Operator has multiple CRDs as well, and they only have one module and there is no separation of concern at the level of data-handler / resource-handler. There will be a number of reconciliation loop regardless, and the main bit about `data-handler` is that it would know something other than deploying Pod etc.

PR #11, opened 2025-10-10
Add minimum kubebuilder and direnv setup

rytswd:
Add directories and files generated by kubebuilder.

NOTE: `internal/controller/` and `cmd/main.go` are purposely left out.

### DIFF
--- a/devshell.nix
+++ b/devshell.nix
@@ -4,6 +4,10 @@ pkgs.mkShell {
   packages = with pkgs; [
     go
     kubebuilder
+    docker
+    docker-buildx
+    kubectl
+    kind
 
     # For some script use cases
     nodejs


devshell.nix
```diff
@@ -3,6 +3,9 @@ pkgs.mkShell {
   # Add build dependencies
   packages = with pkgs; [
     go
+    kubebuilder
```
rcambrj, line 6:
```suggestion
    kubebuilder
    docker
    docker-buildx
    kubectl
    kind
```

`golangci-lint`, `kustomize`, `controller-gen`, and `setup-envtest` appear to live in `bin/` somehow.

Makefile
```diff
@@ -0,0 +1,238 @@
...
+	$(KUSTOMIZE) build config/default | $(KUBECTL) apply -f -
+
+.PHONY: undeploy
+undeploy: kustomize ## Undeploy controller from the K8s cluster specified in ~/.kube/config. Call with ignore-not-found=true to ignore resource not found errors during deletion.
+	$(KUSTOMIZE) build config/default | $(KUBECTL) delete --ignore-not-found=$(ignore-not-found) -f -
+
+##@ Dependencies
+
+## Location to install dependencies to
+LOCALBIN ?= $(shell pwd)/bin
```
rcambrj, line 175: should we add this directory to `.gitignore`?

.envrc
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
```
rcambrj, line 1: oops! thanks for adding this in :)

devshell.nix
rytswd: The kubebuilder way is to have these installed as static binary via Go build pipeline (and thus will be put into the `bin/` directory at project root). I don't think that's necessarily great given Nix can handle this well, but at the same time, I don't want to deviate from how it's meant to be done with kubebuilder. (In that sense, adding `kubebuilder` here is actually wrong, because `kubebuilder init` won't allow devshell or any other file to be present, and even if you try to invoke with absolute path, it somehow reference itself as `kubebuilder` and fails to find the binary, and returns an error.)

golangci-lint is fine to add here, I think. kubebuilder generates thorough (and extremely lengthy) definition, which I didn't include in this PR -- but I decided it's probably better to have it here as well: abed0d6

Makefile
rytswd: I'm surprised why this wasn't added... a26e7ca

devshell.nix
rcambrj: @rytswd I'm confused - from the Makefile, `kubebuilder`, `docker` (and the buildx plugin), `kubectl` and `kind` are all used from `$PATH` (not installed). Only `golangci-lint`, `kustomize`, `controller-gen`, and `setup-envtest` are installed into `bin/` and consumed from that directory.

I wouldn't add the second set into the nix devshell on account of them getting installed by the `make` through that task, but the first set should be OK.

What am I misunderstanding?

rcambrj: Ah nevermind, I see you added these dependencies in 7dbc258d46747ed9a932f890db73c9578633d267, so I think we're on the same page.

rytswd: The only thing is golangci-lint, which I think is nice to have it available

rytswd: But we can do that when we need it

PR #18, opened 2025-10-11
Add utility metadata package for Kubernetes labels

rytswd:
This is a common label handling logic that will be used by many resources.

### DIFF
--- a/pkg/resource-handler/controller/metadata/labels_test.go
+++ b/pkg/resource-handler/controller/metadata/labels_test.go
@@ -1,9 +1,11 @@
-package metadata
+package metadata_test
 
 import (
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
+
+	"github.com/numtide/multigres-operator/pkg/resource-handler/controller/metadata"
 )
 
 func TestBuildStandardLabels(t *testing.T) {
@@ -18,12 +20,12 @@ func TestBuildStandardLabels(t *testing.T) {
 			componentName: "etcd",
 			cellName:      "cell-1",
 			want: map[string]string{
-				LabelAppName:       AppNameMultigres,
-				LabelAppInstance:   "my-etcd-cluster",
-				LabelAppComponent:  "etcd",
-				LabelAppPartOf:     AppNameMultigres,
-				LabelAppManagedBy:  ManagedByMultigres,
-				LabelMultigresCell: "cell-1",
+				"app.kubernetes.io/name":       "multigres",
+				"app.kubernetes.io/instance":   "my-etcd-cluster",
+				"app.kubernetes.io/component":  "etcd",
+				"app.kubernetes.io/part-of":    "multigres",
+				"app.kubernetes.io/managed-by": "multigres-operator",
+				"multigres.com/cell":           "cell-1",
 			},
 		},
 		"empty cellName uses default": {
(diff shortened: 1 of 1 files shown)


pkg/resource-handler/controller/metadata/labels.go
```diff
@@ -0,0 +1,94 @@
+package metadata
+
+import "maps"
+
+// Standard Kubernetes label keys following kubernetes.io conventions.
+//
+// See: https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/
```
fernando-villalba, line 7: This link is helpful here! Looking at it though, I see that there are additional labels defined here:

```yaml
metadata:
  labels:
    app.kubernetes.io/name: mysql
    app.kubernetes.io/instance: mysql-abcxyz
    app.kubernetes.io/version: "5.7.21"
    app.kubernetes.io/component: database
    app.kubernetes.io/part-of: wordpress
    app.kubernetes.io/managed-by: Helm
```

We are not defining `part-of`, or `version` here. Is this by design?

pkg/resource-handler/controller/metadata/labels.go
```diff
@@ -0,0 +1,94 @@
...
+	LabelMultigresCell = "multigres.com/cell"
+)
+
+// BuildStandardLabels builds the standard Kubernetes labels for a Multigres
+// component. These labels are applied to all resources managed by the operator.
+//
+// Parameters:
+//   - resourceName: The name of the custom resource instance (e.g., "my-etcd-cluster")
+//   - componentName: The component type (e.g., "etcd", "gateway", "orch", "pooler")
+//   - cellName: Optional cell name. If empty, no cell label is added.
```
fernando-villalba, line 53: Would this be empty and not set to a default value? I am just wondering, if no cells defined there is still one cell, no?

pkg/resource-handler/controller/metadata/labels.go
rytswd: That's a good question, I think it won't harm to have it added. I will look into this in a bit

pkg/resource-handler/controller/metadata/labels.go
rytswd: You are right; we need to check what the default cell name should be, but I'll add a logic to default to some predefined default when cell information is not provided

pkg/resource-handler/controller/metadata/labels.go
rytswd: This is handled with 2f7e403 and 4ddea7d.

pkg/resource-handler/controller/metadata/labels.go
rytswd: This is handled with ed84b5b and 4ddea7d.

pkg/resource-handler/controller/metadata/labels.go
```diff
@@ -0,0 +1,106 @@
...
+	DefaultCellName = "multigres-global-topo"
+)
+
+// BuildStandardLabels builds the standard Kubernetes labels for a Multigres
+// component. These labels are applied to all resources managed by the operator.
+//
+// Parameters:
+//   - resourceName: The name of the custom resource instance (e.g., "my-etcd-cluster")
+//   - componentName: The component type (e.g., "etcd", "gateway", "orch", "pooler")
+//   - cellName: Optional cell name. If empty, no cell label is added.
```
fernando-villalba, line 53: Maybe remove this comment now we are doing a default?

PR #19, opened 2025-10-11
Add storage utility for PVC

rytswd:
~~Should be reviewed after #18 is merged~~

This is ready for review, simple PVC handling logic


pkg/resource-handler/controller/storage/pvc.go
```diff
@@ -0,0 +1,44 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name: name,
+		},
+		Spec: corev1.PersistentVolumeClaimSpec{
+			AccessModes: []corev1.PersistentVolumeAccessMode{
+				corev1.ReadWriteOnce,
+			},
+			Resources: corev1.VolumeResourceRequirements{
+				Requests: corev1.ResourceList{
+					corev1.ResourceStorage: resource.MustParse(storageSize),
```
fernando-villalba, line 33: No sure here, but I think what will happen here is that it will crash the controller  if a user enters an incorrect size value. Shouldn't we use instead?:

```go
resource.ParseQuantity(storageSize)
```

fernando-villalba: We can resolve this comment as the CRD validation layer will likely take care of this issue, so mustparse should be fine.

PR #15, opened 2025-10-10
Etcd API and CRD

rytswd:

PR #20, opened 2025-10-13
Add etcd controller code

rytswd:
All the implementation logic is in place, as well as test coverage is near 100% test. There is only one function that's not being triggered due to the lack of envtest setup at the moment, and will be probably tackled in a separate PR.

Note that this now uses #15 and #23 for CRDs. Also, this must get the go.mod version reference update for `api` directory.

### DIFF
--- a/pkg/resource-handler/controller/etcd/etcd_controller_test.go
+++ b/pkg/resource-handler/controller/etcd/etcd_controller_test.go
@@ -1,6 +1,7 @@
 package etcd
 
 import (
+	"slices"
 	"testing"
 
 	appsv1 "k8s.io/api/apps/v1"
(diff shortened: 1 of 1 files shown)


pkg/resource-handler/controller/etcd/etcd_controller.go
```diff
@@ -0,0 +1,266 @@
...
+				return fmt.Errorf("failed to create StatefulSet: %w", err)
+			}
+			return nil
+		}
+		return fmt.Errorf("failed to get StatefulSet: %w", err)
+	}
+
+	// Update existing StatefulSet
+	existing.Spec = desired.Spec
+	existing.Labels = desired.Labels
```
fernando-villalba, line 135: I think this should be fine as we would be adding additional labels via the CR itself. I do wonder however if there are many scenarios where companies have automated tooling that adds labels to resources separately, this would effectively overwrite them. I wonder if the best practice is to have that tooling update the CR itself or for us to update this logic so it only overwrites the labels that the operator is supposed to manage, and that's it.

rytswd: That could happen with other mutating webhook affecting running resources -- but I think that's more common with Pod, not at StatefulSet level. We may want to take a note of these and work it out when doing live cluster testing

pkg/resource-handler/controller/etcd/etcd_controller.go
```diff
@@ -0,0 +1,267 @@
...
+
+// SetupWithManager sets up the controller with the Manager.
+// TODO: This is missing test coverage, and will need to use envtest setup.
+func (r *EtcdReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&multigresv1alpha1.Etcd{}).
+		Owns(&appsv1.StatefulSet{}).
+		Owns(&corev1.Service{}).
+		Complete(r)
+}
```
rytswd, line 267: This is the code path that's not yet tested.

pkg/resource-handler/go.mod
```diff
@@ -4,26 +4,63 @@ go 1.25.0
 
 require (
 	github.com/google/go-cmp v0.7.0
+	github.com/numtide/multigres-operator/api v0.0.0-20251010210109-0981dc938d36
```
rytswd, line 7: - [ ] Ensure go.mod pulls in the correct version reference before merging

PR #27, opened 2025-10-14
Switch pipeline to build Go inside Dockerfile

fernando-villalba:
* changes `Containerfile` to `Dockerfile`
* builds Go application inside `Dockerfile`
  * switches intermediate builder to `FROM golang:alpine`
* `build-go` step now only performs linting and tests
* `create-release` step removed

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -24,7 +24,14 @@ COPY . .
 # was called. For example, if we call make docker-build in a local env which has the Apple Silicon M1 SO
 # the docker BUILDPLATFORM arg will be linux/arm64 when for Apple x86 it will be linux/amd64. Therefore,
 # by leaving it empty we can ensure that the container and binary shipped on it will have the same platform.
-RUN CGO_ENABLED=0 GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH} go build -a -o manager cmd/multigres-operator/main.go
+RUN CGO_ENABLED=0 \
+	GOOS=${TARGETOS:-linux} \
+	GOARCH=${TARGETARCH} \
+	go build \
+	-ldflags '-s -w -buildid=' \
+	-trimpath -mod=readonly \
+	-a -o manager \
+	cmd/multigres-operator/main.go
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+WORKDIR /workspace
+# Copy the Go Modules manifests
+COPY go.mod go.mod
+COPY go.sum go.sum
+# cache deps before building and copying source so that we don't need to re-download as much
+# and so that source changes don't invalidate our downloaded layer
+RUN go mod download
+
+# Copy the Go source (relies on .dockerignore to filter)
+COPY . .
```
rcambrj, line 20: let's add .dockerignore with kubebuilder's default `bin/` line

Dockerfile
```diff
@@ -0,0 +1,31 @@
+# Build the manager binary
+FROM golang:1.25 AS builder
```
rcambrj: please add the comment about scanning this intermediate image and keeping this line in sync with the github workflow

.github/workflows/build-and-release.yaml
```diff
@@ -32,8 +32,6 @@ jobs:
         uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
 
       - uses: actions/dependency-review-action@56339e523c0409420f6c2c9a2f4292bbb3c07dd3 # v4.8.0
-        # this action requires a base and head
```
rcambrj, line 35: any reason why comments have been removed throughout this file?

.github/workflows/build-and-release.yaml
rcambrj: it's not necessary to download the artifacts anymore.
in fact, it's not necessary to build them or upload them anymore.

Dockerfile
rcambrj: I know that you've commented the intermediate scanning block in the github workflow, but I think that it'd be prudent to reinstate the comment here, just to keep the two image versions in sync even though the scanning is disabled.

PR #32, opened 2025-10-14
Add sample

rcambrj:
* adds a sample manifest, it brings up etcd!

Screenshot 2025-10-14 at 18 17 55

### DIFF
--- a/config/samples/reference.yaml
+++ b/config/samples/reference.yaml
@@ -1,3 +1,6 @@
+# This sample is intended to represent a complete cluster.
+# Useful as a starting point and as a complete reference for all options.
+
 apiVersion: v1
 kind: Namespace
 metadata:


config/samples/reference.yaml
```diff
@@ -0,0 +1,11 @@
...
+kind: Namespace
+metadata:
+  name: multigres-sample
+---
+apiVersion: multigres.com/v1alpha1
+kind: Etcd
+metadata:
+  namespace: multigres-sample
+  name: sample
+spec: {}
```
rytswd, line 14: The sample configuration should have the detailed spec in here, even if it were to duplicate from the default values. We can have some "extreme minimum" spec just for reference, but this isn't exactly how people should be using the operator spec, so I would just be cautious how to name this.

rytswd (approved):
OK to go in as is for now, but definitely need rewording / renaming, as I think many would check out sample config like this to have a quick grasp of the project.

config/samples/reference.yaml
rcambrj: Agreed. I hadn't planned for this to be a minimum sample, let's plan to grow this with all the specs as a reference. Is that agreeable? Would you prefer a different name?

rytswd: I'm OK with that, but we need a clear indication that this is still WIP. Can you rename, or simply add a comment, like this?

```suggestion
# TODO: This is a test spec for dev testing, we will need to update this with more fields
apiVersion: multigres.com/v1alpha1
kind: Etcd
metadata:
  namespace: multigres-sample
  name: sample
spec: {}
```

rcambrj: Agreed! I'll add a comment. But I don't think that such a TODO can ever be considered complete (because when is a software project ever complete? 😄 ) so if it's alright with you, I'll add:

```
# This sample is intended to represent a complete cluster.
# Useful as a starting point and as a complete reference for all options.
```

...and rename to `reference.yaml`

If you feel it necessary, might I encourage opening a github issue to remind ourselves about filling in all the options?

rytswd: The TODO comments can be listed from most IDEs, and with a new project like this, I thought it would help us spot them before making a release. I'm OK with your comment as well, the whole point is to clarify this isn't meant for general setup.

Also, I'm fine with having an issue. Before the official release, we will need to work on every single configuration exposed to users, and full documentation. We are still far away from it -- having an issue may help as a reminder, but we would probably still go through every piece of code / config / doc anyways.

rcambrj: > to clarify this isn't meant for general setup.

why not?

PR #39, opened 2025-10-16
DRAFT: Add cilium integration doc

rytswd:


teaglebuilt:
I think there is a new sig for handling this problem that cilium is using https://multicluster.sigs.k8s.io/concepts/multicluster-services-api/. I imagine we should be prepared for connecting clusters across hybrid environments.

PR #46, opened 2025-10-16
create local cluster deployment path

rcambrj:
main goal: fixes `make deploy`

* removes example (non-existent) role references
* adds `minikube` to devshell
* adds `make minikube-load`
* renames default container image to something which kubernetes won't try to pull from network


fernando-villalba (commented):
Similar to this one: https://github.com/numtide/multigres-operator/pull/45/files

PR #52, opened 2025-10-20
Add multiorch controller

rytswd:


pkg/resource-handler/controller/multiorch/deployment.go
```diff
@@ -0,0 +1,87 @@
...
+	multigresv1alpha1 "github.com/numtide/multigres-operator/api/v1alpha1"
+	"github.com/numtide/multigres-operator/pkg/resource-handler/controller/metadata"
+)
+
+const (
+	// ComponentName is the component label value for multiorch resources
+	ComponentName = "multiorch"
+
+	// DefaultReplicas is the default number of multiorch replicas
+	DefaultReplicas int32 = 1
```
fernando-villalba, line 21: I've looked at a couple of different operators, and the defaults are usually defined at the API spec level, with  [a global defaults file](https://github.com/planetscale/vitess-operator/blob/main/pkg/apis/planetscale/v2/defaults.go)  and then [defaults per spec file](https://github.com/planetscale/vitess-operator/blob/main/pkg/apis/planetscale/v2/vitesscluster_defaults.go). Perhaps we would want to adopt a similar approach? What do you think?

rytswd: The global default file is useful when the same definition is used in multiple places -- but they would have to be public variables for other packages to pull in. I'm not sure if we want to have so many const / var definitions in a single place, where it's actually only consumed by a single package.

I think I would drop the default replicas, or make it so that it's explicitly 0. I know this is covered by the type definition -- I may also change the default there.

fernando-villalba: What about defining the default per spec then? It does feel like the defaults should be defined within the API code, not the operator code. Up to you though.

pkg/resource-handler/controller/multiorch/multiorch_controller.go
```diff
@@ -0,0 +1,258 @@
...
+		return fmt.Errorf("failed to get Service: %w", err)
+	}
+
+	// Update existing Service
+	existing.Spec.Ports = desired.Spec.Ports
+	existing.Spec.Selector = desired.Spec.Selector
+	existing.Spec.Type = desired.Spec.Type
+	existing.Labels = desired.Labels
+	existing.Annotations = desired.Annotations
+	if err := r.Update(ctx, existing); err != nil {
```
fernando-villalba, line 183: Could we consider using SSA by default instead of this where applicable? 

Something like this:

```go
import (
    "sigs.k8s.io/controller-runtime/pkg/client"
    metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
)

// Build a Service object with desired state (partial)
applyOpts := []client.PatchOption{
    client.ForceOwnership,               // Take ownership of fields you specify
    client.FieldOwner("multigres-operator"), 
}
desiredApply := &corev1.Service{
    ObjectMeta: metav1.ObjectMeta{
        Name:        existing.Name,
        Namespace:   existing.Namespace,
        Labels:      desired.Labels,
        Annotations: desired.Annotations,
    },
    Spec: corev1.ServiceSpec{
        Ports:    desired.Spec.Ports,
        Selector: desired.Spec.Selector,
        Type:     desired.Spec.Type,
    },
}

// Apply patch using Server-Side Apply
if err := r.Patch(ctx, desiredApply, client.Apply, applyOpts...); err != nil {
    return fmt.Errorf("failed to apply Service: %w", err)
}
```

PR #53, opened 2025-10-23
Clean up Makefile to match with the multi module setup

rytswd:
Instead of using kubebuilder commands directly, we can now use `make` commands.

Notable support is:
- `make kind-up`: Create kind cluster
- `make manifests`: Create manifests based on `/api/` directory
- `make generate`: Code generate for DeepCopy, etc., based on `/api/` directory
- `make modules-xyz`: Update all modules

And there are a few more scripts that support all the Go modules in one go.

There is also some version details added to the operator initalisation.

### DIFF
--- a/devshell.nix
+++ b/devshell.nix
@@ -18,6 +18,7 @@ pkgs.mkShell rec {
 
   # Add environment variables
   env = {
+    # NOTE: This is used with `make setup-envtest`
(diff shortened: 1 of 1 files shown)


rytswd:
Creating new resources is something I should look into. Probably not so much for the controller, that one is only rudimentary setup that isn't all that useful IMO. I'll follow-up in a separate PR.

PR #54, opened 2025-10-23
New API Spec

fernando-villalba:
The API spec is now ready to review. You can use this design as reference to check the API definition:

https://github.com/numtide/multigres-operator/blob/multigres-cluster-cr-design/plans/phase-1/final-cr-definition.md

Please note that the pipeline failed due to the KubeBuilder generated controller files, which we could remove if needed, or use as a reference to code forward.


api/v1alpha1/cell_types.go
```diff
@@ -0,0 +1,200 @@
...
+	Replicas *int32 `json:"replicas,omitempty"`
+
+	// Affinity defines the pod's scheduling constraints.
+	// +optional
+	Affinity *corev1.Affinity `json:"affinity,omitempty"`
+
+	// Resources defines the compute resource requirements.
+	// +optional
+	corev1.ResourceRequirements `json:"resources,omitempty"`
+}
```
rytswd, line 89: Because this is used by multiple resources, let's move this to a separate file

internal/controller/cell_controller.go
rytswd, line 1: Can you please remove all the code from `internal/controller`, as we will not be using them (and they are basically useless templates as is).

api/v1alpha1/tablegroup_types.go
rytswd, line 1: I think this was meant to be renamed to "tablegroup_types.go"?

rytswd (commented):
I didn't go through the entire code yet, but just left a few quick ones for now.

PR #64, opened 2025-11-21
Remove use of underscore to fix CRD load error

rytswd:


rytswd:
A quick adjustment for old CRDs that are to be removed. A safe change, will go ahead with the merge.

PR #65, opened 2025-11-21
Fix lint error

rytswd:
This seems to be blocking the latest CRD spec to be picked up


rytswd:
Going ahead with the merge of this minor change

PR #66, opened 2025-11-21
Add testutil for resource watching and declarative testing

rytswd:
With the new utility and go-cmp as dependency, we can gather all the events from kube-apiserver, and simply compare the desired state and match against what we get from etcd.

This allows fully parallel test to run, and no TestMain juggling needed.


rytswd:
image
Test coverage is still missing, but for now, we have got most of the logic correct. Also, given this is a testing only utility, we will look at filling the rest of missing test cases later in a separate PR.

Going ahead with the merge on this one.

PR #68, opened 2025-11-22
Update pool slice to map to match with spec

rytswd:
This is what it looks like based on https://github.com/numtide/multigres-operator/blob/multigres-cluster-cr-design/plans/phase-1/api-design/multigres-operator-api-v1alpha1-design.md#child-cr-shard

``` yaml
kind: Shard
metadata:
  name: "production-db-orders-tg-0"
  namespace: example
  labels:
     # ...
  ownerReferences:
    - apiVersion: multigres.com/v1alpha1
      # ...
spec:
  shardName: "0"

  images:
    # ...
  
  multiOrch:
    image: "multigres/multigres:latest"
    # ....

  pools:
    primary:
      cell: "us-east-1a" 
      type: "readWrite"
      replicas: 2
      storage:
        # ...
```

It used to be a slice, and now it is using as a map, which gives easier mapping and suffix usage for the resource names.

### DIFF
--- a/api/v1alpha1/shard_types.go
+++ b/api/v1alpha1/shard_types.go
@@ -37,9 +37,9 @@ type ShardSpec struct {
 	MultiOrch MultiOrchSpec `json:"multiOrch,omitempty"`
 
 	// Pools defines the different pools of pods for this shard (e.g., replicas, read-only).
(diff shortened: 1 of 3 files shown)


rytswd:
This is a minor change that's needed for the controller implementation. Going ahead with the merge.

PR #69, opened 2025-11-26
Admission controller doc

fernando-villalba:
A document detailing defaults, controller logic and where to apply it.

### DIFF
--- a/plans/phase-1/admission-controller-and-defaults.md
+++ b/plans/phase-1/admission-controller-and-defaults.md
@@ -24,8 +24,14 @@ This layered approach uses the most efficient tool for each task, from basic fie
 A review of classic operators like `vitess-operator` highlights a major user-facing weakness: applying complex defaults *in-memory* during reconciliation.
 
 * **The "Classic" Problem:** The `vitess-operator` avoids webhooks entirely. It uses simple OpenAPI v3 CRD defaults (Level 1) for static values but applies all complex, dynamic defaults (like images and resource requests) via Go code *after* fetching the object (Level 2).
-* **The User-Facing Confusion:** Because these Level 2 defaults are not persisted back to `etcd`, a user running `kubectl get` on a CR sees a spec that does **not** match the functional state the operator is enforcing. This is confirmed by a 6-year-old `TODO` in their code explicitly mentioning the lack of mutating webhook support in Operator SDK as a blocker.
-* **Our Motivation:** This "invisible" default pattern is unacceptable for `multigres-operator`. Our API relies on a sophisticated **4-Level Override Chain** (Inline -> Cluster Default -> Namespace Default -> Operator Hardcoded). We *must* make sure our manifests are fully declarative and the user sees the complete, resolved CR with `kubectl get`. A **mutating webhook** is the correct solution, as it applies all defaults and persists them to `etcd` *before* the reconciler ever sees the object.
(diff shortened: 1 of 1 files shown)


plans/phase-1/admission-controller-and-defaults.md
```diff
@@ -0,0 +1,453 @@
...
+
+  * Implement a **mutating webhook** to persist complex defaults (e.g., image tags, resource blocks) into the `MultigresCluster` spec, solving the "invisible defaults" problem.
+  * Implement a **stateful validating webhook** for rules that require cluster-awareness (querying other objects), such as:
+      * Preventing the deletion of a `DeploymentTemplate` that is in use by any `MultigresCluster`.
+      * Validating that a referenced `DeploymentTemplate` exists upon `MultigresCluster` creation or update.
+  * Utilize **CRD-Embedded CEL (`XValidation`)** for all advanced *stateless, object-only* validation (e.g., mutual exclusion, odd numbers, required sub-fields).
+  * Utilize **`ValidatingAdmissionPolicy` (CEL)** for *stateless, context-aware* validation that needs access to `oldObject.status` or `request.userInfo`.
+  * Provide clear, actionable error messages at admission time.
+  * Support multiple certificate management strategies (cert-manager and self-contained).
+  * Keep webhook logic in the `data-handler` module.
```
teaglebuilt: what is the distinction of the data-handler and why here instead of resource-handler?

plans/phase-1/admission-controller-and-defaults.md
```diff
@@ -0,0 +1,453 @@
...
+
+### 3\. Certificate Management
+
+Since we **must** run a webhook server (Layer 4), we must manage TLS certificates. We will support three strategies.
+
+#### Strategy A: cert-manager (Recommended for Production)
+
+  * **Pros:** Automatic certificate generation, renewal, and CA bundle injection. Industry standard.
+  * **Cons:** Adds a runtime dependency on `cert-manager`.
+  * **Implementation:** Ship `Certificate` and `Issuer` manifests in `config/webhook/cert-manager`.
```
teaglebuilt: how would we prevent the webhook from registering before the certs are ready? For example, the api server tries to call the webhook (MultiCluster CREATE/DELETE) before certs are generated

fernando-villalba: @teaglebuilt 


We sidestep the issue by enforcing a strict order of operations in `main.go`.

Here is how:

1.  First, we check if `tls.crt` and `tls.key` already exist in the cert directory (e.g., mounted by cert-manager). If they do, we skip internal generation entirely and just use those files.
2.   If *no* certs are found, we initialize our internal generator and run `rotator.Bootstrap(ctx)`. This is a blocking call. The operator won't proceed past this line until the certs are generated, written to disk, and the `MutatingWebhookConfiguration` is patched with the new CA bundle.
3.  The webhook server (port 9443) is managed by `mgr`. Since we haven't called `mgr.Start()` yet, the binary isn't even listening on that port. If the API server tried to call us during this phase, it would just get a connection refused, not a TLS error.
4.  On top of that, we use a `readyz` check. Kubernetes won't send traffic to the Pod's Service until the manager is fully up and running, which only happens *after* that bootstrap phase succeeds.

I hope that answers your question! The code is already implemented in main if you want to go take a look.

PR #70, opened 2025-11-26
feat(api): refactor v1alpha1 to parent/child model with scoped templates

fernando-villalba:
Completely redesigns the API based on the latest design doc, moving from individual component CRDs to a single "MultigresCluster" source of truth supported by reusable templates and read-only child resources.

Key Changes:
- **MultigresCluster**: Re-implemented as the Root CR. strictly ordered structs and added "templateDefaults" and "overrides" logic.
- **Templates**: Added `CoreTemplate`, `CellTemplate`, and `ShardTemplate` types to support reusable configuration.
- **Shared Types**: Created `shared_types.go` to centralize common structs (`StatelessSpec`, `ComponentConfig`, `StorageSpec`) and prevent circular deps.
- **TopoServer**: Refactored `toposerver_types.go` to include `EtcdSpec` and `GlobalTopoServerRef`, acting as the authoritative source for topology config.
- **Shard**: Moved `MultiOrchSpec`, `PoolSpec`, and `ShardImages` into `shard_types.go` to localize domain-specific logic.
- **Children**: Updated `Cell`, `TableGroup`, and `Shard` to reflect their status as read-only, fully resolved child resources.

Refactoring & Cleanup:
- Deleted obsolete component types: `deploymenttemplate`, `etcd`, `multigateway`, and `multiorch`.
- Applied strict "definition-before-use" ordering and added ASCII headers to all API files for better readability.
- Standardized naming conventions (e.g., using `Spec` over `Config` for managed resources).


api/v1alpha1/cell_types.go
```diff
@@ -17,148 +17,68 @@ limitations under the License.
...
-	// +optional
-	RegisterCell bool `json:"registerCell,omitempty"`
-
-	// PruneTablets instructs the controller to prune old tablets from the topology.
-	// +optional
-	PruneTablets bool `json:"pruneTablets,omitempty"`
+// TopologyReconciliation defines flags for the cell controller.
+type TopologyReconciliation struct {
+	RegisterCell bool `json:"registerCell"`
+	PruneTablets bool `json:"pruneTablets"`
```
rytswd: Is this the right terminology? Or just taken from Vitess?

fernando-villalba: You are right! I just changed it both [here](https://github.com/numtide/multigres-operator/pull/70/commits/2a15e2c5694bf9258d052a5e85b29397131ecf76) and in the design spec!

api/v1alpha1/toposerver_types.go
```diff
@@ -22,79 +22,146 @@ import (
...
+
+// ============================================================================
+// TopoServer Component Specs
+// ============================================================================
+//
+// These components are not directly used in the formation of this child CR,
+// but they are used to configure the toposerver via the MultigresCluster
+
+// EtcdSpec defines the configuration for a managed Etcd cluster.
+type EtcdSpec struct {
```
rytswd, line 69: Can you tell me why we have both "TopoServerSpec" and "EtcdSpec", both with identical set of fields?  The comment about being used for MultigresCluster isn't clear "WHY" that has to be the case.

api/v1alpha1/common_types.go
```diff
@@ -0,0 +1,95 @@
...
+// ============================================================================
+// Shared Configuration Structs
+// ============================================================================
+//
+// These structs are used across multiple resources (Cluster, Templates, Children)
+// to ensure consistency in configuration shapes.
+
+// StatelessSpec defines the desired state for a scalable, stateless component
+// like MultiAdmin, MultiOrch, or MultiGateway.
+type StatelessSpec struct {
```
rytswd, line 32: We don't have any volume setup here, is that correct?

api/v1alpha1/common_types.go
```diff
@@ -0,0 +1,95 @@
...
+type ComponentConfig struct {
+	// Spec defines the inline configuration.
+	// +optional
+	Spec *StatelessSpec `json:"spec,omitempty"`
+
+	// TemplateRef refers to a CoreTemplate to load configuration from.
+	// +optional
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=63
+	TemplateRef string `json:"templateRef,omitempty"`
```
rytswd: This says it's about "CoreTemplate". Is that true? Looking at the template, it does look like this is only used for CoreTemplate. At that point, this is not really shared / common. I don't see a reason why we need this nested struct definition, and we can simply define this in CoreTemplate directly?

api/v1alpha1/common_types.go
rytswd, line 1: This is just my preference, but because we have "SHARD", can we rename this file to `common_types.go` instead?

fernando-villalba: This is a phenomenal suggestion. Yes, I will rename so it's easier not to confuse them. Changed on 691086d

api/v1alpha1/common_types.go
fernando-villalba: No, because it's stateless.

api/v1alpha1/common_types.go
fernando-villalba: Another great catch! Changed on 1d02795

api/v1alpha1/toposerver_types.go
fernando-villalba: This is also a phenomenal catch. This happened because I changed the name of `managedSpec` to `etcd` to accommodate potential future use-cases when we may have more than just etcd as a topology server. However it is clear that my implementation was half cooked because even having two different types here, the ending child CR is not polymorphic, it's just a copy of etcdSpec.

Before I change anything here, let me propose two solutions to you and you tell me which one you like best.


## 1. We get rid of `etcdSpec` and just use `topoServerSpec`. 
 
The drawback here is that it isn't very future proof if we add additional keyvalue stores in the future but it slightly simplifies the spec. The advantage is that it is simpler to do and wouldn't require altering the design document.


## 2. We make `topoServerSpec` "polymorphic" 

We keep `etcdSpec` as it is but we add a nesting key under topoServerSpec like so:

```go
// api/v1alpha1/toposerver_types.go

// EtcdSpec defines the configuration for a managed Etcd cluster.
// Shared by both Parent and Child.
type EtcdSpec struct {
	// Image is the Etcd container image.
	// +optional
	// +kubebuilder:validation:MinLength=1
	// +kubebuilder:validation:MaxLength=512
	Image string `json:"image,omitempty"`

	// Replicas is the desired number of etcd members.
	// +kubebuilder:validation:Minimum=1
	// +optional
	Replicas *int32 `json:"replicas,omitempty"`

	// Storage configuration for Etcd data.
	// +optional
	Storage StorageSpec `json:"storage,omitempty"`

	// Resources defines the compute resource requirements.
	// +optional
	Resources corev1.ResourceRequirements `json:"resources,omitempty"`
}

// TopoServerSpec defines the desired state of TopoServer (Child CR).
// Now polymorphic!
// +kubebuilder:validation:XValidation:rule="has(self.etcd)",message="must specify 'etcd' configuration"
type TopoServerSpec struct {
	// Etcd defines the configuration if using Etcd.
	// +optional
	Etcd *EtcdSpec `json:"etcd,omitempty"`

    // Future-proofing example:
    // Consul *ConsulSpec `json:"consul,omitempty"`
}
```


I am leaning towards option 2, even though it requires a small change to the design spec, but let me know your thoughts.

rytswd: I like the option 2 as well. etcd is only a specific implementation for this KV setup. I've seen many solutions using different sort of KVs, and even inventing their own for their use cases. I think this change is simple enough (I don't call it polymorphic, though, it's just a wrapper struct)

api/v1alpha1/common_types.go
rytswd: Stateless does not necessary mean there has to be no volume, though? Volume could be used as a readonly mount. I'm thinking about some secret mounting? It may be a stretch, but if the data written to the volume is only for diagnostics purposes, I would still consider such process stateless

fernando-villalba: I say we leave it as it is for now for the sake of simplicity, if in the future there is a use-case for volumes we can change it.

api/v1alpha1/toposerver_types.go
fernando-villalba: I've made the changes on f8572cb

api/v1alpha1/common_types.go
rytswd: OK noted, thanks

PR #75, opened 2025-12-13
Adjust RBAC setup for full deployment

rytswd:
~~Once this RBAC setup is in place, with #62 and #74 are merged, I could confirm all the logic to work with simple:~~

As both PRs above are merged, I could confirm the following would bring up each component correctly. Because of the missing images, though, components fail to come up, but that's expected.

```console
❯ make kind-up
❯ make kind-deploy

# Ensure kubeconfig is pulled out
❯ kind get kubeconfig --name multigres-operator-dev > /tmp/kind-kubeconfig.yaml
❯ kubectl --kubeconfig /tmp/kind-kubeconfig.yaml apply -f config/samples/child-resources/

# Monitor processes coming up
❯ k9s --kubeconfig /tmp/kind-kubeconfig.yaml

# Clean up
❯ make kind-down
```

### DIFF
--- a/config/samples/child-resources/kind-cell.yaml
+++ b/config/samples/child-resources/kind-cell.yaml
@@ -23,6 +23,17 @@ spec:
         cpu: 200m
         memory: 256Mi
 
+  # MultiOrch deployment - cell orchestration
+  multiorch:
+    replicas: 1
+    resources:
+      requests:
+        cpu: 50m
+        memory: 64Mi
+      limits:
+        cpu: 200m
+        memory: 128Mi
+
   # Reference to the global topology server
   globalTopoServer:
     rootPath: /multigres


config/samples/child-resources/kind-cell.yaml
rytswd, line 1: Copying a comment from the commit message:

This may need to be more clear that they are only for testing and not meant to be used as they are.

config/samples/child-resources/kind-cell.yaml
```diff
@@ -0,0 +1,73 @@
...
+    resources:
+      requests:
+        cpu: 50m
+        memory: 128Mi
+      limits:
+        cpu: 200m
+        memory: 256Mi
+
+  # MultiOrch deployment - cell orchestration
+  multiorch:
```
fernando-villalba, line 27: Placing multiorch here is not correct, unless we are still working with an old version of the spec here for now

rytswd: Ah thanks  for the catch, corrected with 11eac79

rytswd: Actually, this removal now breaks the deployment. The initial `cell` spec we introduced to this repo assumed multiorch as non-optional requirements. We need to update the entire API definitions based on the spec document separately, so I just reverted the above commit with c8ef74e, and we can tackle that separately.

fernando-villalba (approved):

PR #78, opened 2025-12-15
Multigres controller draft

fernando-villalba:


pkg/cluster-handler/controller/multigrescluster/multigrescluster_controller.go
```diff
@@ -0,0 +1,456 @@
...
+	}
+
+	if cluster.ObjectMeta.DeletionTimestamp.IsZero() {
+		if !controllerutil.ContainsFinalizer(cluster, finalizerName) {
+			controllerutil.AddFinalizer(cluster, finalizerName)
+			if err := r.Update(ctx, cluster); err != nil {
+				return ctrl.Result{}, err
+			}
+		}
+	} else {
```
rytswd: When the deletion timestamp is there, we have to complete the reconciliation. Otherwise it would try to update the "deleting" resource. Once the resource becomes "deleting", you cannot "undelete", and thus this must be corrected.

Also, that means you will not need the `else` keyword, because once it hits the if statement, it MUST return, and there is no need to else. This goes with the idiomatic Go code of error path being indented, so that the main path stays non-indented.

fernando-villalba: Changed on 190410c

I apologize it is a big change, I had to also change tests and I was struggling to get it to 100%, so far I am on 99.3%:

ok       github.com/numtide/multigres-operator/pkg/cluster-handler/controller/multigrescluster   (cached)        coverage: 99.3% of statements
ok      github.com/numtide/multigres-operator/pkg/cluster-handler/controller/tablegroup (cached)        coverage: 100.0% of statements

PR #80, opened 2025-12-17
Adding multigrescluster and tablegroup controller

fernando-villalba:
I will be closing [PR 78 ](https://github.com/numtide/multigres-operator/pull/78) in favour of this one as it is using all the latest version of the API and code in main, making for a cleaner merge.

### DIFF
--- a/.github/workflows/_reusable-test-coverage.yaml
+++ b/.github/workflows/_reusable-test-coverage.yaml
@@ -93,6 +93,11 @@ jobs:
           echo "Previous test coverage: Not found"
           fi
 
+          ## Setup Workspace for Tooling
+          # This forces Go to look at local directories for modules instead of trying to fetch them from online.
+          go work init
+          go work use -r .
+
           ## Step 2. Check new coverage
           # Get into directory so that go tool cover can work
           go tool cover -func=coverage/combined.out > /tmp/coverage.txt


pkg/cluster-handler/controller/multigrescluster/multigrescluster_controller.go
```diff
@@ -0,0 +1,509 @@
...
+
+// Reconcile reads that state of the cluster for a MultigresCluster object and makes changes based on the state read
+// and what is in the MultigresCluster.Spec.
+//
+// +kubebuilder:rbac:groups=multigres.com,resources=multigresclusters,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=multigres.com,resources=multigresclusters/status,verbs=get;update;patch
+// +kubebuilder:rbac:groups=multigres.com,resources=multigresclusters/finalizers,verbs=update
+// +kubebuilder:rbac:groups=multigres.com,resources=coretemplates;celltemplates;shardtemplates,verbs=get;list;watch
+// +kubebuilder:rbac:groups=multigres.com,resources=cells;tablegroups;toposervers,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=apps,resources=deployments,verbs=get;list;watch;create;update;patch;delete
```
rytswd, line 42: FYI, these markers won't be read by default. This should be corrected in Makefile. (For Resource Handler use cases, I moved the RBAC spec out from the controller temporarily.)

pkg/cluster-handler/controller/multigrescluster/multigrescluster_controller.go
```diff
@@ -0,0 +1,509 @@
...
+		if controllerutil.ContainsFinalizer(cluster, finalizerName) {
+			if err := r.checkChildrenDeleted(ctx, cluster); err != nil {
+				return ctrl.Result{}, err
+			}
+			controllerutil.RemoveFinalizer(cluster, finalizerName)
+			if err := r.Update(ctx, cluster); err != nil {
+				return ctrl.Result{}, fmt.Errorf("failed to remove finalizer: %w", err)
+			}
+		}
+		return ctrl.Result{}, nil
```
rytswd: In my resource handler code, I adopted a separate function for this deletion handling, something like 
```go
if !cluster.ObjectMeta.DeletionTimestamp.IsZero() {
  return handleDelete(...)
}
```

That way, it is clear that any logic related to deletion would return early, and can be separated from the success code path. I think it helps with the code clarity a lot.

pkg/cluster-handler/controller/multigrescluster/multigrescluster_controller.go
```diff
@@ -0,0 +1,509 @@
...
+		return fmt.Errorf("failed to list toposervers: %w", err)
+	}
+	if len(ts.Items) > 0 {
+		return fmt.Errorf("toposervers still exist")
+	}
+
+	return nil
+}
+
+func (r *MultigresClusterReconciler) reconcileGlobalComponents(ctx context.Context, cluster *multigresv1alpha1.MultigresCluster, resolver *TemplateResolver) error {
```
rytswd: Can you make two separate functions for the resolution logic here, one for topo server, and another for multi admin? Because it changes the context back and forth between topo server and multi admin, I find it difficult to follow the code.

pkg/cluster-handler/controller/multigrescluster/multigrescluster_controller_test.go
```diff
@@ -0,0 +1,1339 @@
...
+					},
+				},
+			},
+		},
+	}
+
+	errBoom := errors.New("boom")
+
+	tests := []struct {
+		name            string
```
rytswd: Can you update the setup so that this is a map based setup, rather than slices? Like:

```go
tests := map[string]struct{
...
}
```

That way, you cannot have a test with the same name, and name should be explicitly given to each test to make it clear what the test is about.

pkg/cluster-handler/controller/multigrescluster/multigrescluster_controller_test.go
```diff
@@ -0,0 +1,1339 @@
...
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+
+	multigresv1alpha1 "github.com/numtide/multigres-operator/api/v1alpha1"
+	"github.com/numtide/multigres-operator/pkg/testutil"
+)
+
+func TestMultigresClusterReconciler_Reconcile(t *testing.T) {
```
rytswd: Please make sure to add `t.Parallel()` for all tests, unless it is absolutely necessary to run in order

pkg/cluster-handler/controller/multigrescluster/multigrescluster_controller_test.go
```diff
@@ -0,0 +1,1339 @@
...
+				WithObjects(tc.existingObjects...).
+				WithStatusSubresource(&multigresv1alpha1.MultigresCluster{}, &multigresv1alpha1.Cell{}, &multigresv1alpha1.TableGroup{})
+			baseClient := clientBuilder.Build()
+
+			var finalClient client.Client
+			if tc.failureConfig != nil {
+				finalClient = testutil.NewFakeClientWithFailures(baseClient, tc.failureConfig)
+			} else {
+				finalClient = baseClient
+			}
```
rytswd, line 1023: I think this can be simplified without any else

```suggestion
			finalClient := client.Client(baseClient)
			// Wrap with failure injection if configured
			if tc.failureConfig != nil {
				finalClient = testutil.NewFakeClientWithFailures(baseClient, tc.failureConfig)
			}
```

pkg/cluster-handler/controller/multigrescluster/multigrescluster_controller_test.go
```diff
@@ -0,0 +1,1339 @@
...
+			name: "Create: Adds Finalizer",
+			cluster: func() *multigresv1alpha1.MultigresCluster {
+				c := baseCluster.DeepCopy()
+				c.Finalizers = nil // Explicitly remove finalizer to test addition
+				return c
+			}(),
+			existingObjects: []client.Object{},
+			expectError:     false,
+			validate: func(t *testing.T, c client.Client) {
+				ctx := context.Background()
```
rytswd: Don't use `context.Background()`, that's the old way, it should be `t.Context()`

pkg/cluster-handler/controller/multigrescluster/multigrescluster_controller_test.go
```diff
@@ -0,0 +1,1339 @@
...
+				updatedCluster := &multigresv1alpha1.MultigresCluster{}
+				_ = c.Get(ctx, types.NamespacedName{Name: clusterName, Namespace: namespace}, updatedCluster)
```
rytswd: We should never ignore error, even in these validation logic. If error found, we should be returning early with `t.Fatal`

pkg/cluster-handler/controller/multigrescluster/multigrescluster_controller_test.go
```diff
@@ -0,0 +1,1339 @@
...
+	errBoom := errors.New("boom")
+
+	tests := []struct {
+		name            string
+		cluster         *multigresv1alpha1.MultigresCluster
+		existingObjects []client.Object
+		failureConfig   *testutil.FailureConfig
+		setupFunc       func(client.Client)
+		expectError     bool
+		validate        func(t *testing.T, c client.Client)
```
rytswd: Anywhere you take `*testing.T` as an argument as a function, we should be using `testing.TB` instead. `testing.T` is specific to `TestXyz` functions, and `testing.B` is for benchmark testing. Using `testing.TB` allows the same code to be used for benchmark testing.

pkg/cluster-handler/controller/multigrescluster/multigrescluster_controller_test.go
```diff
@@ -0,0 +1,1339 @@
...
+	}{
+		// ---------------------------------------------------------------------
+		// Success Scenarios
+		// ---------------------------------------------------------------------
```
rytswd: Because we have so many test cases, I would even suggest splitting the success cases and failure cases into separate functions.

pkg/cluster-handler/controller/multigrescluster/multigrescluster_controller_test.go
```diff
@@ -0,0 +1,1339 @@
...
+				},
+			},
+		},
+	}
+
+	errBoom := errors.New("boom")
+
+	tests := []struct {
+		name            string
+		cluster         *multigresv1alpha1.MultigresCluster
```
rytswd: Can we rename this field to `multigrescluster`, as it's quite confusing whether this is about Kubernetes cluster or Multigres cluster?
(I'm not going to add suggestion here, as you'd probably want to use LSP refactor instead)

pkg/cluster-handler/controller/multigrescluster/multigrescluster_controller_test.go
```diff
@@ -0,0 +1,1339 @@
...
+		// ---------------------------------------------------------------------
+		// Success Scenarios
+		// ---------------------------------------------------------------------
+		{
+			name: "Create: Adds Finalizer",
+			cluster: func() *multigresv1alpha1.MultigresCluster {
+				c := baseCluster.DeepCopy()
+				c.Finalizers = nil // Explicitly remove finalizer to test addition
+				return c
+			}(),
```
rytswd: This is a very strange line, especially how this pattern is used throughout the test cases. If we are to just have one field, we should just do `cluster: someClusterResource`. If we need more complex logic, we should do `cluster: createClusterResource` where createClusterResource takes `testing.TB`.

If we need more adjustments before feeding it into the reconciliation logic, we can have another `preReconcileUpdate: func(t testing.TB, mc *multigresv1alpha1.MultigresCluster) *multigresv1alpha1.MultigresCluster`.

pkg/cluster-handler/controller/multigrescluster/multigrescluster_controller_test.go
```diff
@@ -0,0 +1,1339 @@
...
+	}
+
+	errBoom := errors.New("boom")
+
+	tests := []struct {
+		name            string
+		cluster         *multigresv1alpha1.MultigresCluster
+		existingObjects []client.Object
+		failureConfig   *testutil.FailureConfig
+		setupFunc       func(client.Client)
```
rytswd: Is setupFunc actually used?

pkg/cluster-handler/controller/multigrescluster/multigrescluster_controller.go
fernando-villalba: Changed on 7e52849

pkg/cluster-handler/controller/multigrescluster/multigrescluster_controller.go
fernando-villalba: Changed on 48a1ec0

pkg/cluster-handler/controller/multigrescluster/multigrescluster_controller_test.go
fernando-villalba: Changed on c86eb45

pkg/cluster-handler/controller/multigrescluster/multigrescluster_controller_test.go
fernando-villalba: Changed on 9d8f7ce

pkg/cluster-handler/controller/multigrescluster/multigrescluster_controller_test.go
fernando-villalba: Changed on 60b96bb

pkg/cluster-handler/controller/multigrescluster/multigrescluster_controller_test.go
fernando-villalba: Changed on 8658528

pkg/cluster-handler/controller/multigrescluster/multigrescluster_controller_test.go
fernando-villalba: Changed on e21681f

pkg/cluster-handler/controller/multigrescluster/multigrescluster_controller_test.go
fernando-villalba: Changed on b4f9fdb

pkg/cluster-handler/controller/multigrescluster/multigrescluster_controller_test.go
fernando-villalba: Changed on 508c7ed

pkg/cluster-handler/controller/multigrescluster/multigrescluster_controller_test.go
fernando-villalba: Changed on c661e22

pkg/cluster-handler/controller/multigrescluster/multigrescluster_controller_test.go
fernando-villalba: Changed on 55ad5ac

pkg/cluster-handler/controller/multigrescluster/multigrescluster_controller_test.go
fernando-villalba: Changed on a04a7f2

pkg/cluster-handler/controller/multigrescluster/multigrescluster_controller_test.go
```diff
@@ -0,0 +1,1310 @@
...
+
+	tests := map[string]struct {
+		multigrescluster   *multigresv1alpha1.MultigresCluster
+		existingObjects    []client.Object
+		preReconcileUpdate func(testing.TB, *multigresv1alpha1.MultigresCluster)
+		setupFunc          func(testing.TB, client.Client)
+		validate           func(testing.TB, client.Client)
+	}{
+		"Create: Adds Finalizer": {
+			multigrescluster: baseCluster.DeepCopy(),
```
rytswd: Looks like this is always started with `baseCluster.DeepCopy()`. At that point, there is no need of making these explicit, and simply do something like

```go
mc := basecCluster.DeepCopy()
if tc.multigrescluster != nil {
    mc = tc.multigrescluster
}
```

(And it may be even worth removing the entry entirely from the table field definitions

pkg/cluster-handler/controller/multigrescluster/multigrescluster_controller_test.go
```diff
@@ -0,0 +1,1310 @@
...
+	_ = appsv1.AddToScheme(scheme)
+	_ = corev1.AddToScheme(scheme)
+
+	coreTpl, cellTpl, shardTpl, baseCluster, clusterName, namespace, finalizerName := setupFixtures()
+
+	tests := map[string]struct {
+		multigrescluster   *multigresv1alpha1.MultigresCluster
+		existingObjects    []client.Object
+		preReconcileUpdate func(testing.TB, *multigresv1alpha1.MultigresCluster)
+		setupFunc          func(testing.TB, client.Client)
```
rytswd: I think this should be removed, I don't see any code using if I'm not mistaken?

pkg/cluster-handler/controller/multigrescluster/multigrescluster_controller_test.go
```diff
@@ -0,0 +1,1310 @@
...
+			},
+		},
+		"Create: Independent Templates (Topo vs Admin)": {
+			multigrescluster: func() *multigresv1alpha1.MultigresCluster {
+				c := baseCluster.DeepCopy()
+				c.Spec.TemplateDefaults.CoreTemplate = "" // clear default
+				c.Spec.GlobalTopoServer.TemplateRef = "topo-core"
+				c.Spec.MultiAdmin.TemplateRef = "admin-core"
+				return c
+			}(),
```
rytswd: This should be using preReconcileUpdate hook instead

pkg/cluster-handler/controller/multigrescluster/multigrescluster_controller_test.go
```diff
@@ -0,0 +1,1310 @@
...
+				deploy := &appsv1.Deployment{}
+				if err := c.Get(ctx, types.NamespacedName{Name: clusterName + "-multiadmin", Namespace: namespace}, deploy); err != nil {
+					t.Fatal("MultiAdmin not created")
+				}
+			},
+		},
+		"Create: MultiOrch Placement Defaulting": {
+			multigrescluster: baseCluster.DeepCopy(),
+			preReconcileUpdate: func(t testing.TB, c *multigresv1alpha1.MultigresCluster) {
+				c.Spec.TemplateDefaults.ShardTemplate = "" // Ensure no default template interferes
```
rytswd, line 267: This makes me think -- if the ShardTemplate is removed from the spec like this, we should be able to just define the ShardTemplate resource into the cluster all the time in the cluster. That would mean you can default to have `existingObjects: []client.Object{coreTpl, cellTpl, shardTpl}` as a part of test setup within `t.Run`, and not having to specify any objects unless you need something else specifically.

pkg/cluster-handler/controller/multigrescluster/multigrescluster_controller_test.go
```diff
@@ -0,0 +1,1310 @@
...
+				// Verify Wiring
+				cell := &multigresv1alpha1.Cell{}
+				if err := c.Get(ctx, types.NamespacedName{Name: clusterName + "-zone-a", Namespace: namespace}, cell); err != nil {
+					t.Fatal("Expected Cell 'zone-a' to exist")
+				}
+
+				expectedAddr := clusterName + "-global-topo-client." + namespace + ".svc:2379"
+				if diff := cmp.Diff(expectedAddr, cell.Spec.GlobalTopoServer.Address); diff != "" {
+					t.Errorf("Wiring Bug! Cell has wrong Topo Address mismatch (-want +got):\n%s", diff)
+				}
```
rytswd: Sorry I didn't notice this earlier, but this is very strange. If it's a single field, you should be just doing equality check. `cmp.Diff` helps you diff between two objects, and gives you a nice diff view of which field is different, etc. 

Let's keep this a simple equality check for some fields we want to verify. The actual declarative comparison should happen in integration_test.go where we can make use of cmp.Diff.

pkg/cluster-handler/controller/multigrescluster/multigrescluster_controller_test.go
```diff
@@ -0,0 +1,1310 @@
...
+			existingObjects: []client.Object{cellTpl, shardTpl}, // No Core Template
+			validate: func(t testing.TB, c client.Client) {
+				// Verify Cell got empty topo address
+				cell := &multigresv1alpha1.Cell{}
+				_ = c.Get(t.Context(), types.NamespacedName{Name: clusterName + "-zone-a", Namespace: namespace}, cell)
+				if diff := cmp.Diff("", cell.Spec.GlobalTopoServer.Address); diff != "" {
+					t.Errorf("Expected empty topo address mismatch (-want +got):\n%s", diff)
+				}
+			},
+		},
```
rytswd: Why do we have this test? I'm just trying to think what this is specifically testing, as I can see missing Core template is also tested with "Create: External Topo with Empty Endpoints"

pkg/cluster-handler/controller/multigrescluster/multigrescluster_controller_test.go
```diff
@@ -0,0 +1,1310 @@
...
+					Spec:       multigresv1alpha1.CellSpec{Name: "zone-a"},
+				}
+				if err := c.Create(ctx, cell); err != nil {
+					t.Fatalf("failed to create cell: %v", err)
+				}
+				cell.Status.Conditions = []metav1.Condition{{Type: "Available", Status: metav1.ConditionTrue}}
+				if err := c.Status().Update(ctx, cell); err != nil {
+					t.Fatalf("failed to update cell status: %v", err)
+				}
+			},
```
rytswd: This can be declarative object spec as a part of `existingObjects` I think?

pkg/cluster-handler/controller/multigrescluster/multigrescluster_controller_test.go
```diff
@@ -0,0 +1,1310 @@
...
+					if controllerutil.ContainsFinalizer(updated, finalizerName) {
+						t.Error("Finalizer was not removed")
+					}
+				}
+			},
+		},
+		"Error: Object Not Found (Clean Exit)": {
+			multigrescluster: baseCluster.DeepCopy(),
+			existingObjects:  []client.Object{},
+		},
```
rytswd: Should this test case be here? If this is a "success" test, we should rename the test

pkg/cluster-handler/controller/multigrescluster/multigrescluster_controller_test.go
```diff
@@ -0,0 +1,1310 @@
...
+				WithObjects(tc.existingObjects...).
+				WithStatusSubresource(&multigresv1alpha1.MultigresCluster{}, &multigresv1alpha1.Cell{}, &multigresv1alpha1.TableGroup{})
+			baseClient := clientBuilder.Build()
+
+			var finalClient client.Client
+			finalClient = baseClient
+
+			if tc.setupFunc != nil {
+				tc.setupFunc(t, baseClient)
+			}
```
rytswd: Immediately assigning isn't idiomatic in Go, so it should be like this


```suggestion
			finalClient := baseClient
			if tc.setupFunc != nil {
				tc.setupFunc(t, baseClient)
			}
```

pkg/cluster-handler/controller/multigrescluster/multigrescluster_controller_test.go
```diff
@@ -0,0 +1,1310 @@
...
+
+			if !strings.Contains(name, "Object Not Found") {
+				check := &multigresv1alpha1.MultigresCluster{}
+				err := baseClient.Get(t.Context(), types.NamespacedName{Name: tc.multigrescluster.Name, Namespace: tc.multigrescluster.Namespace}, check)
+				if apierrors.IsNotFound(err) {
+					if err := baseClient.Create(t.Context(), tc.multigrescluster); err != nil {
+						t.Fatalf("failed to create initial cluster: %v", err)
+					}
+				}
+			}
```
rytswd, line 602: We shouldn't have a special logic based on the test case name, which could be updated (and it is so easy to not use the same casing, etc.).

Also, some code correction even if were to add this logic somehow.
- we shouldn't be using `baseClient`, it should be `finalClient`
- when `err` is returned, it is only checked with `apierrors.IsNotFound` - but if it returned "any" error, we should do early t.Fatal

I don't even know what this is trying to achieve, though...?

pkg/cluster-handler/controller/multigrescluster/multigrescluster_controller_test.go
```diff
@@ -0,0 +1,1310 @@
...
+					Name:      tc.multigrescluster.Name,
+					Namespace: tc.multigrescluster.Namespace,
+				},
+			}
+
+			_, err := reconciler.Reconcile(t.Context(), req)
+
+			if err != nil {
+				t.Errorf("Unexpected error from Reconcile: %v", err)
+			}
```
rytswd, line 619: It's idiomatic to do an immediate error check without extra line
```suggestion
			_, err := reconciler.Reconcile(t.Context(), req)
			if err != nil {
				t.Errorf("Unexpected error from Reconcile: %v", err)
			}
```

pkg/cluster-handler/controller/multigrescluster/multigrescluster_controller_test.go
```diff
@@ -0,0 +1,1310 @@
...
+)
+
+// setupFixtures helper returns a fresh set of test objects to ensure isolation between test functions.
+func setupFixtures() (
+	*multigresv1alpha1.CoreTemplate,
+	*multigresv1alpha1.CellTemplate,
+	*multigresv1alpha1.ShardTemplate,
+	*multigresv1alpha1.MultigresCluster,
+	string, string, string,
+) {
```
rytswd, line 36: Any private helper function should take `testing.TB`

```suggestion
// setupFixtures helper returns a fresh set of test objects to ensure isolation between test functions.
func setupFixtures(tb testing.TB) (
	*multigresv1alpha1.CoreTemplate,
	*multigresv1alpha1.CellTemplate,
	*multigresv1alpha1.ShardTemplate,
	*multigresv1alpha1.MultigresCluster,
	string, string, string,
) {
	tb.Helper()
```

rytswd: And with that additional arg, the comment can be probably dropped.

pkg/cluster-handler/controller/multigrescluster/multigrescluster_controller_test.go
```diff
@@ -0,0 +1,1310 @@
...
+func TestMultigresClusterReconciler_Reconcile_Failure(t *testing.T) {
+	t.Parallel()
+
+	scheme := runtime.NewScheme()
+	_ = multigresv1alpha1.AddToScheme(scheme)
+	_ = appsv1.AddToScheme(scheme)
+	_ = corev1.AddToScheme(scheme)
+
+	coreTpl, cellTpl, shardTpl, baseCluster, clusterName, namespace, finalizerName := setupFixtures()
+	errBoom := errors.New("boom")
```
rytswd, line 639: What's this? If we are injecting some special error for testing, we should describe it more clearly.

pkg/cluster-handler/controller/multigrescluster/multigrescluster_controller_test.go
fernando-villalba: Changed on b8a050f

pkg/cluster-handler/controller/multigrescluster/multigrescluster_controller_test.go
fernando-villalba: Changed on fbd230a

pkg/cluster-handler/controller/multigrescluster/multigrescluster_controller_test.go
fernando-villalba: Changes on cd48659

pkg/cluster-handler/controller/multigrescluster/multigrescluster_controller_test.go
fernando-villalba: Changed on 2dc9e48

pkg/cluster-handler/controller/multigrescluster/multigrescluster_controller_test.go
fernando-villalba: Changed on abf2a39

pkg/cluster-handler/controller/multigrescluster/multigrescluster_controller_test.go
fernando-villalba: Change on fc18a00

pkg/cluster-handler/controller/multigrescluster/multigrescluster_controller_test.go
fernando-villalba: Changed on 9861b16

pkg/cluster-handler/controller/multigrescluster/multigrescluster_controller_test.go
fernando-villalba: changed on bb9d2b8

pkg/cluster-handler/controller/multigrescluster/multigrescluster_controller_test.go
fernando-villalba: Changed on 2562d8c

pkg/cluster-handler/controller/multigrescluster/multigrescluster_controller_test.go
fernando-villalba: Changed on 8a14d5f

pkg/cluster-handler/controller/multigrescluster/multigrescluster_controller_test.go
fernando-villalba: changed on 3694e40

pkg/cluster-handler/controller/multigrescluster/multigrescluster_controller_test.go
fernando-villalba: This is a sentinel error used to inject failures into the mock client and verify the controller handles them correctly. I used 'boom' as a placeholder, but I agree it's unclear. Since we have made loads of changes to this PR already and we need to make progress I won't be changing this now. But I'll make a note for next iteration.

PR #85, opened 2025-12-22
fix(api): allow optional Cell TopoServer and adjust CEL validation limits

fernando-villalba:
- Change `Cell.Spec.TopoServer` to a pointer (`*LocalTopoServerSpec`). This allows the field to be omitted (nil) when using the global default topology server, resolving validation errors where the empty struct failed the "oneOf" requirement (etcd vs external).

- Adjusted validation limits due to CEL budget issues during integration tests. These limits may need to be adjusted as we go.

### DIFF
--- a/api/v1alpha1/common_types.go
+++ b/api/v1alpha1/common_types.go
@@ -45,13 +45,13 @@ type StatelessSpec struct {
 
 	// PodAnnotations are annotations to add to the pods.
 	// +optional
-	// +kubebuilder:validation:MaxProperties=48
+	// +kubebuilder:validation:MaxProperties=20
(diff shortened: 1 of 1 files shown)


api/v1alpha1/multigrescluster_types.go
```diff
@@ -272,7 +272,7 @@ type ShardOverrides struct {
 
 	// Pools overrides. Keyed by pool name.
 	// +optional
-	// +kubebuilder:validation:MaxProperties=32
+	// +kubebuilder:validation:MaxProperties=8
```
rytswd, line 275: Just curious, where is this number coming from?

fernando-villalba: From my desperation and frugality to keep everything under CEL budget 😭 

Don't take any of these limits as gospel, my guess is that we will need to adjust them further as we go along.

PR #87, opened 2025-12-24
Fix API mismatch for resource-handler

rytswd:
This ensures all the API changes are incorporated into resource-handler. Also, some commandline arguments have been corrected.

### DIFF
--- a/pkg/resource-handler/controller/shard/shard_controller_internal_test.go
+++ b/pkg/resource-handler/controller/shard/shard_controller_internal_test.go
@@ -399,6 +399,7 @@ func TestUpdateStatus_MultiOrch(t *testing.T) {
 		expectError     bool
 		expectOrchReady bool
 		setupClient     func(*testing.T, *runtime.Scheme, *multigresv1alpha1.Shard) client.Client
+		customShard     *multigresv1alpha1.Shard // Optional: override default shard
 	}{
 		"GetError": {
 			expectError: true,
@@ -449,6 +450,28 @@ func TestUpdateStatus_MultiOrch(t *testing.T) {
 				},
 			},
 		},
+		"NotFound": {
+			expectError:     false,
+			expectOrchReady: false,
+			setupObjects:    []client.Object{}, // No MultiOrch Deployment - will get NotFound
+		},
+		"NoCellsInMultiOrchOrPools": {
+			expectError:     false,
+			expectOrchReady: false,
+			setupObjects:    []client.Object{},
+			customShard: &multigresv1alpha1.Shard{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "test-shard",
+					Namespace: "default",
+				},
+				Spec: multigresv1alpha1.ShardSpec{
+					MultiOrch: multigresv1alpha1.MultiOrchSpec{
+						Cells: []multigresv1alpha1.CellName{}, // Empty
+					},
+					Pools: map[string]multigresv1alpha1.PoolSpec{}, // Empty
+				},
+			},
+		},
 	}
 
 	for name, tc := range tests {
@@ -457,17 +480,21 @@ func TestUpdateStatus_MultiOrch(t *testing.T) {
 			_ = multigresv1alpha1.AddToScheme(scheme)
 			_ = appsv1.AddToScheme(scheme)
 
-			shard := &multigresv1alpha1.Shard{
-				ObjectMeta: metav1.ObjectMeta{
-					Name:      "test-shard",
-					Namespace: "default",
-				},
-				Spec: multigresv1alpha1.ShardSpec{
-					MultiOrch: multigresv1alpha1.MultiOrchSpec{
-						Cells: []multigresv1alpha1.CellName{"zone1"},
+			// Use custom shard if provided, otherwise use default
+			shard := tc.customShard
+			if shard == nil {
+				shard = &multigresv1alpha1.Shard{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      "test-shard",
+						Namespace: "default",
 					},
-					Pools: map[string]multigresv1alpha1.PoolSpec{},
-				},
+					Spec: multigresv1alpha1.ShardSpec{
+						MultiOrch: multigresv1alpha1.MultiOrchSpec{
+							Cells: []multigresv1alpha1.CellName{"zone1"},
+						},
+						Pools: map[string]multigresv1alpha1.PoolSpec{},
+					},
+				}
 			}
 
 			var fakeClient client.Client


rytswd:
I haven't done go.work + main.go testing in my local kind. I'll want to do that before merging this -- will work on that a bit later.

pkg/resource-handler/controller/shard/labels.go
```diff
@@ -11,22 +13,45 @@ import (
 func buildPoolLabels(
```
fernando-villalba: Looks like you kept this but you are not using it anymore?

pkg/resource-handler/controller/shard/pool_service.go
```diff
@@ -16,16 +16,17 @@ const (
...
 	poolName string,
-	poolSpec multigresv1alpha1.ShardPoolSpec,
+	cellName string,
+	poolSpec multigresv1alpha1.PoolSpec,
 	scheme *runtime.Scheme,
 ) (*corev1.Service, error) {
-	name := buildPoolName(shard.Name, poolName)
-	labels := buildPoolLabels(shard, poolName, poolSpec)
+	name := buildPoolNameWithCell(shard.Name, poolName, cellName)
+	labels := buildPoolLabelsWithCell(shard, poolName, cellName, poolSpec)
```
fernando-villalba, line 29: You changed this but the declaration for buildPoolLabels remains and it is unused.

pkg/resource-handler/controller/shard/labels.go
```diff
@@ -11,22 +13,45 @@ import (
...
-		cellName = metadata.DefaultCellName
+
+	// Build comma-separated list of cells for the label
+	cellNames := make([]string, len(poolSpec.Cells))
+	for i, cell := range poolSpec.Cells {
+		cellNames[i] = string(cell)
 	}
+	cellLabel := metadata.DefaultCellName
+	if len(cellNames) > 0 {
+		cellLabel = strings.Join(cellNames, ",")
```
fernando-villalba: This will cause a runtime error during resource creation?

The function `metadata.AddCellLabel` (defined in `pkg/resource-handler/controller/metadata/labels.go`  blindly assigns this string value to the label map without sanitization:

```go
// pkg/resource-handler/controller/metadata/labels.go

// AddCellLabel adds the cell label to the provided labels map.
func AddCellLabel(labels map[string]string, cellName string) map[string]string {
	labels[LabelMultigresCell] = cellName
	return labels
}

```

Because this propagates the raw comma-separated string (e.g., `"us-east-1a,us-east-1b"`) directly to the Kubernetes object `metadata.labels`, the API server will reject the request.

Kubernetes label values must follow strict syntax rules (RFC 1123):

> *Must be 63 characters or less, and must be empty or begin and end with an alphanumeric character ([a-z0-9A-Z]) with dashes (-), underscores (_), dots (.), and alphanumerics between.*

**Commas are strictly forbidden in label values.** If we need to represent multiple cells, we cannot use a single label key for this list.

pkg/resource-handler/controller/shard/shard_controller.go
```diff
@@ -60,16 +60,28 @@ func (r *ShardReconciler) Reconcile(
...
 
-	// Reconcile MultiOrch Deployment
-	if err := r.reconcileMultiOrchDeployment(ctx, shard); err != nil {
-		logger.Error(err, "Failed to reconcile MultiOrch Deployment")
-		return ctrl.Result{}, err
+	// Reconcile MultiOrch - one Deployment and Service per cell
+	cells := shard.Spec.MultiOrch.Cells
+	if len(cells) == 0 {
+		return ctrl.Result{}, fmt.Errorf(
+			"MultiOrch has no cells specified - cannot deploy without cell information",
```
fernando-villalba: This is a deviation from the design document:

"The Operator will deploy one instance of this Deployment into EVERY Cell listed in 'cells'. If 'cells' is empty, it defaults to all cells where pools are defined."

pkg/resource-handler/controller/shard/shard_controller.go
```diff
@@ -316,9 +351,9 @@ func (r *ShardReconciler) updateStatus(
 	}
 
 	// Update status fields
-	shard.Status.TotalPods = totalPods
-	shard.Status.ReadyPods = readyPods
-	shard.Status.ObservedGeneration = shard.Generation
```
fernando-villalba, line 321: I noticed you removed these for cell and shard, why?

This field is critical for Kubernetes clients (and other controllers) to determine if the current Status reflects the most recent Spec. Without it, clients cannot distinguish between "status hasn't updated yet" and "status is up-to-date but nothing changed."

pkg/resource-handler/controller/shard/shard_controller.go
```diff
@@ -316,9 +351,9 @@ func (r *ShardReconciler) updateStatus(
...
+	shard.Status.PoolsReady = (totalPods > 0 && totalPods == readyPods)
+	// TODO: Add OrchReady status check when MultiOrch deployment is implemented
```
fernando-villalba: Since the deployment logic is landing in this PR, why not add this already?

pkg/resource-handler/controller/shard/containers.go
```diff
@@ -68,28 +63,34 @@ func buildPostgresContainer(
...
 
 	// TODO: Add remaining command line arguments:
-	// --topo-global-server-addresses (needs global topo server ref in ShardSpec)
-	// --topo-global-root (needs global topo server ref in ShardSpec)
 	// --pooler-dir, --grpc-socket-file, --log-level, --log-output, --hostname, --service-map
+	// --pgbackrest-stanza, --connpool-admin-password, --socket-file
 
 	args := []string{
+		"multipooler", // Subcommand
 		"--http-port", "15200",
```
fernando-villalba, line 85: Not a big deal now. No need to fix at this point, but at some point we may want to make these configurable and default them.

fernando-villalba:
You don't need to address this comment now but I thought I would mention it so we can be prepare soon.

To follow up with the resolver module I created in preparation for the webhook (to avoid where`kubectl get` doesn't show the actual values being used, among other things). The `pkg/resolver` should now serve as a Single Source of Truth for defaults, usable by both the Mutating Webhook (when enabled) and the Reconciler (when disabled).

Here are the specific areas we will need to refactor to align with that design:

**1. Centralizing defaults (Single Source of Truth)**
Currently, this PR defines local constants like `DefaultMultiGatewayImage` in `multigateway.go`. My branch also defines `DefaultMultiGatewayImage` in `pkg/resolver/defaults.go`.

* **The Risk:** If these drift, the Webhook might default the Parent CR to version `v1.0`, but if a user omits a field on the Child CR, this controller might default it to `v1.1`.
* **The Fix:** Once merged, we should delete the constants in `resource-handler` and import them directly from `pkg/resolver` to guarantee consistency.

**2. The "Read-Only Child" & Explicit Specs**
The design states that Child CRs (Cell, Shard) are "read-only" and strictly owned by the `MultigresCluster`.

* **Current Logic:** Your code creates defensive defaults (e.g., `if spec.Image == "" { image = Default }`).
* **Target Logic:** The `MultigresCluster` controller will use `resolver` to calculate all these values *before* creating the Child CR. Therefore, the Child CR's `.Spec` should arrive fully populated.
* **Future Change:** We should move away from "Just-in-Time" defaulting inside these child controllers. Instead, they should ideally trust that the Spec is populated (or error out if it's missing, alerting us that the Parent Controller failed to resolve defaults). This ensures that `kubectl get cell` always matches exactly what is running.

**3. Webhook Enablement/Disablement**
The `resolver` package is designed to be called explicitly by the Parent Controller if the Webhook is disabled. 
```go
// Inside MultigresCluster Reconcile loop
resolvedCluster := resolver.Resolve(cluster)
// Now create Child CRs using resolvedCluster.Spec...
```
This works regardless of whether a webhook ran previously. If the webhook ran, cluster comes in fully populated. If not, Resolve populates it in memory.

This means the `Cell` CR will *always* receive explicit values from the Parent, regardless of whether a webhook ran. This makes the defensive defaulting in this PR redundant in the long term.

pkg/resource-handler/controller/shard/labels.go
rytswd: Indeed, this was left by mistake, fixed with 4aaa207

pkg/resource-handler/controller/shard/pool_service.go
rytswd: The same as above, fixed with 4aaa207

pkg/resource-handler/controller/shard/labels.go
rytswd: Thanks for the note, I was under impression that commas were allowed. Given the implementation change with a single "cell" information to be passed in rather than multiple "cells" for shard creation, this logic was removed with 4aaa207.

pkg/resource-handler/controller/shard/shard_controller.go
rytswd: Hmm, I see. Can we consider explicitly assigning the list of cells from top level resource (MultigresCluster) to the Shard resources instead? Because the templates resolution results in a very explicit spec for child resources, this "defaulting" logic for child resources feels a bit mismatched.

For now, I changed the logic to follow the documented spec with 105fac6

pkg/resource-handler/controller/shard/shard_controller.go
rytswd: Sorry, this was simply because I tried to work out the other parts of the logic first, and was meant to be corrected. This was corrected with a6f7537

pkg/resource-handler/controller/shard/shard_controller.go
rytswd: Indeed, I didn't pay too much attention for the status handling as I thought I could get back to it easily, but this was simple enough to add. Corrected with a6f7537

pkg/resource-handler/controller/shard/containers.go
rytswd: Hmmm, but for what use case do you expect? I just can't see any reason for such flexibility / configurability.

pkg/resource-handler/controller/shard/shard_controller.go
```diff
@@ -189,33 +201,53 @@ func (r *ShardReconciler) reconcileMultiOrchService(
...
 ) error {
-	// Reconcile pool StatefulSet
-	if err := r.reconcilePoolStatefulSet(ctx, shard, poolName, poolSpec); err != nil {
-		return fmt.Errorf("failed to reconcile pool StatefulSet: %w", err)
+	// Pools must have cells specified
+	if len(poolSpec.Cells) == 0 {
+		return fmt.Errorf(
+			"pool %s has no cells specified - cannot deploy without cell information",
+			poolName,
+		)
```
rytswd, line 216: Similar to how MultiOrch was set to fail, Pool setup also requires the explicit list of cells right now. Should we correct this?

pkg/resource-handler/controller/shard/containers.go
fernando-villalba: Then that's okay, just leave it as it is. We shouldn't add any more features and configuration than required, this would be easy enough to change later anyway.

pkg/resource-handler/controller/shard/shard_controller.go
fernando-villalba: This is good, cells is a requirement here so no need for you to change it.

PR #92, opened 2025-12-31
Introduce a separate resolver package

rytswd:
This supports template and default handling, which can be used for both controller logic as well as validation with webhook.

NOTE: This is based on the implementation from #90, with commit moved around to ensure clean merge.


rytswd:
Anyone can delete the branch with editor access 🫠

PR #95, opened 2026-01-01
Fix base image to multigres:main instead of multigres:latest

rytswd:


rytswd:
Going ahead with the merge for this really simple change.

PR #159, opened 2026-01-29
refactor(resources): enforce immutable identity selectors

fernando-villalba:
Updated resource selectors (Deployments, StatefulSets, Services) to use a strict "Identity Filter" instead of matching all object labels.

- Introduced `metadata.GetSelectorLabels()` to filter labels down to a stable identity set (Component, Instance, Cluster, Database, Shard, Pool, Cell).
- Excluded mutable metadata (e.g., `managed-by`, `zone`, `region`) from selectors to prevent "immutability traps" where updating operator metadata would require recreating StatefulSets.
- Updated all controllers (Cluster, Shard, Cell, Topo) to apply this filter.


fernando-villalba:
Merging first to bump the module versions

PR #178, opened 2026-02-03
feat(observability): implement comprehensive events and strict recording

fernando-villalba:
The operator previously lacked visibility into state transitions, child resource updates, and implicit defaulting, making debugging difficult. Additionally, defensive nil-checks for the EventRecorder masked uninitialized dependencies in the test suite.

* Refactored `PopulateClusterDefaults` to return decision strings for implicit configuration
* Added `ImplicitDefault` events to MultigresCluster controller
* Implemented `Applied` events in Shard, Cell, TableGroup, and TopoServer controllers for all child resources
* Added `PhaseChange` events to track lifecycle transitions across all controllers
* Updated all controller test suites to initialize `FakeRecorder` to prevent panics

Significantly enhances operator observability via standard Kubernetes events and improves code quality by enforcing strict dependency invariants.


fernando-villalba:
Need to merge to bump the version

PR #183, opened 2026-02-03
fix(controller): enforce strict observed generation checks

fernando-villalba:
Controllers currently can report false positive "Ready" states momentarily due to cache lag immediately after patching child resources (read-after-write inconsistency).

* Added `ObservedGeneration` field to `CellStatus`, `ShardStatus`, and `TopoServerStatus` definitions.
* Updated `updateStatus` in resource controllers to explicitly sync `ObservedGeneration` with the metadata generation.
* Refactored condition building logic to strictly require matching generations before marking resources as "Ready".

Prevents status flapping during updates and guarantees accurate deployment state for automation pipelines.

### DIFF
--- a/api/v1alpha1/cell_types.go
+++ b/api/v1alpha1/cell_types.go
@@ -107,6 +107,10 @@ type TopologyReconciliation struct {
 
 // CellStatus defines the observed state of Cell.
 type CellStatus struct {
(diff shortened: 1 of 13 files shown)


fernando-villalba:
Merging to bump modules

PR #278, opened 2026-02-23
feat(api): gate localTopoServer behind CEL validation

fernando-villalba:
The localTopoServer field exists in CellInlineSpec and CellTemplateSpec but the Cell controller does not act on it. Upstream multigres has not validated per-cell topology servers in practice, so setting the field would silently do nothing.

- Add XValidation CEL rule on CellInlineSpec and CellTemplateSpec rejecting localTopoServer with "not yet supported" message
- Regenerate CRDs to include the new validation rules
- Document the feature gap, existing scaffolding, upstream status, and future implementation steps in implementation-notes.md

Prevents silent misconfiguration; the CEL markers can be removed and CRDs regenerated when upstream support lands.

### DIFF
--- a/api/v1alpha1/celltemplate_types.go
+++ b/api/v1alpha1/celltemplate_types.go
@@ -25,6 +25,7 @@ import (
 // ============================================================================
 
 // CellTemplateSpec defines reusable config for Cell components (Gateway, LocalTopo).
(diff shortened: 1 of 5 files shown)


fernando-villalba:
Merging now, but module versions need to be updated.

PR #281, opened 2026-02-23
fix(rbac): add watch verb for storageclasses

fernando-villalba:
The resolver's hasDefaultStorageClass uses the cached client to list StorageClasses, but controller-runtime's informer cache requires both list and watch permissions to set up its ListWatch reflector.

Having only list caused the StorageClass informer to fail with "storageclasses.storage.k8s.io is forbidden" on every watch attempt, flooding operator logs with errors.


fernando-villalba:
Merging now, updating modules later

PR #291, opened 2026-02-26
feat(e2e): add GH Action setup with comment event

rytswd:


rytswd:
/e2e

Testing, not sure if it works when the GitHub Action definition is still in a branch

PR #294, opened 2026-02-26
fix(e2e): ensure to add go.sum entry for testutil

rytswd:


rytswd:
Going ahead with the simple fix

PR #311, opened 2026-02-26
fix(action): make runs sequential to handle timeout

rytswd:


rytswd:
Going ahead with the merge on this simple one

PR #328, opened 2026-03-01
docs(plans): add data-handler refactor proposal

fernando-villalba:
The operator uses two controllers per CRD (resource-handler and data-handler), creating finalizer races, broken drain ordering, and topology registration at the wrong hierarchy level.

- Add plans/phase-1/data-handler-refactor-proposal.md
- Propose converting data-handler controllers to utility packages
- Move topo registration to MultigresCluster (parent registers children pattern)
- Introduce PendingDeletion annotation for safe-before-delete
- Implement prune-on-reconcile for poolers, databases, and cells
- Eliminate all finalizers via pre-deletion cleanup
- Document phased migration strategy across 7 phases
- Add questions for upstream multigres team on shard CRUD and pruning safety

Enables individual shard deletion, cell removal, and future multi-shard/multi-database support.

### DIFF
--- a/plans/phase-1/data-handler-refactor-proposal.md
+++ b/plans/phase-1/data-handler-refactor-proposal.md
@@ -1110,59 +1110,6 @@ refactor.
 
 ---
 
-## 11. Questions for the Multigres Team
-
-Before implementing, the following questions should be clarified with the
-upstream multigres team:
-
-### 11.1 Shard Registration in Topology
-
-The upstream `topoclient` currently has no `CreateShard`/`DeleteShard` API.
-Shards exist implicitly under the database path
-(`databases/{db}/{tablegroup}/{shard}/`) and are only materialised when
-`LockShard()` creates the lock path.
-
-Our operator currently registers databases via `CreateDatabase()`, and the
-shard's existence is implicit — tracked through the Shard CR at the
-Kubernetes level and the pooler entries that reference a `ShardKey`.
-
-**Questions:**
-
-1. Is there a plan to add explicit shard CRUD to the `topoclient` for
-   multi-shard support (e.g. `CreateShard`, `GetShardNames`,
-   `DeleteShard`)? If so, what data would a shard record contain (key
-   range, serving state, primary tablet, etc.)?
-
-2. In the meantime, is it correct for the operator to continue registering
-   only the database via `CreateDatabase()` and let shards remain implicit?
-   Or should we be creating shard directory nodes explicitly?
-
-### 11.2 Operator-Side Pruning
-
-We plan to implement a prune-on-reconcile pattern where the operator:
-
-- **Shard controller:** lists pooler entries in topo via
-  `GetMultiPoolersByCell()`, compares to running pods, and calls
-  `UnregisterMultiPooler()` for any stale entries.
-- **MultigresCluster controller:** lists databases/cells in topo via
-  `GetDatabaseNames()`/`GetCellNames()`, compares to the spec, and calls
-  `DeleteDatabase()`/`DeleteCell()` for entries no longer in the spec.
-
-**Questions:**
-
-1. Are there any concerns with the operator calling `UnregisterMultiPooler()`
-   to remove stale pooler entries, beyond what the drain state machine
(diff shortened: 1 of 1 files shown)


plans/phase-1/data-handler-refactor-proposal.md
```diff
@@ -0,0 +1,1142 @@
...
+4. With pooler pruning, stale entries self-heal if the parent recreates the
+   shard (i.e. the shard is still in the TableGroup spec) and the new
+   controller runs its prune logic.
+
+### 6.3 "Make It Safe, Then Delete" Pattern
+
+Instead of trying to drain during deletion (when the CR is dying and
+controllers are racing), adopt this pattern:
+
+**For shard deletion:**
```
Verolop, line 681: this looks good. How does TableGroup know when to come back and delete after it sets PendingDeletion?

plans/phase-1/data-handler-refactor-proposal.md
```diff
@@ -0,0 +1,1142 @@
...
+
+Mature production operators managing stateful distributed databases achieve
+**zero finalizers** by pushing all topo registration to the parent level and
+doing cleanup before deletion, not during it.
+
+---
+
+## 4. Topology Registration: What's Wrong and How to Fix It
+
+### 4.1 The Self-Registration Anti-Pattern
```
Verolop, line 378: the current data-handler creates a new topo connection per reconcile. After the merge, the Shard reconciler needs topo for four phases (PodRoles, drain, backup health, pooler prune). Are we opening a connection per phase or once at the top?

plans/phase-1/data-handler-refactor-proposal.md
```diff
@@ -0,0 +1,1142 @@
...
+Since there are no external users yet, this can be done as a single
+refactoring effort. Recommended order:
+
+1. **Create topo utility packages** (Phase 1) — smallest scope, no behaviour
+   change, proves the extraction pattern.
+2. **Merge Cell controllers** (Phase 3) — smaller controller, proves the
+   merge pattern. **Note:** at this point the merged Cell controller still
+   contains topo registration code temporarily. This is a deliberate stepping
+   stone — the topo logic moves to MultigresCluster in the next step.
+3. **Move topo registration to MultigresCluster** (Phase 2) — fixes the
```
Verolop, line 974: steps 2 and 3 could be one PR. The Cell controller is tiny (MultiGateway Deployment + Service + topo registration). Doing them separately means the merged Cell controller temporarily has topo code that moves to MultigresCluster in the next PR.

fernando-villalba: Oh I will probably do it all in one PR, but split it into a few commits, not sure how many yet. I will probably know tomorrow.

plans/phase-1/data-handler-refactor-proposal.md
fernando-villalba: Currently the data-handler opens up to 3 separate topo connections per reconcile (one per function that needs topo). After the merge, the single reconciler opens one connection at the top of Reconcile() and passes it to all phases. The sequential pipeline makes this straightforward, the current dual-controller architecture made it harder because the functions were written independently across separate controllers that couldn't share state.

plans/phase-1/data-handler-refactor-proposal.md
fernando-villalba: 1. TableGroup controller sees a shard is no longer in the spec → sets PendingDeletion annotation on the Shard CR

1. Shard controller sees the PendingDeletion annotation on its next reconcile → calls handlePendingDeletion() → drains all pods using the existing drain state machine (pod by pod: requested → draining → acknowledged → ready-for-deletion) → once all pods are drained, sets ReadyForDeletion=True on the Shard's status conditions

1. TableGroup controller gets re-reconciled (triggered by the Shard status change via its watch) → sees ReadyForDeletion=True → calls Delete() on the Shard CR → CR deleted immediately, no finalizer to block

The key thing is that steps 2 and 3 involve different controllers. The Shard controller does the work, the TableGroup controller makes the decision to delete. The parent waits for the child to confirm it's safe before pulling the trigger.

PR #462, opened 2026-04-13
feat(cell): add first-class TLS support for multigateway via cert-manager

haritabh17:
## Summary

- Adds `certCommonName` field to `MultigresClusterSpec` and `CellSpec` CRDs
- Cell controller creates a cert-manager `Certificate` (using `supabase-issuer` ClusterIssuer) when `certCommonName` is set
- Multigateway deployment conditionally mounts the TLS secret and passes `--pg-tls-cert-file` / `--pg-tls-key-file` flags
- TLS is fully gated on `certCommonName` — clusters without cert-manager are unaffected

Closes MUL-341

## Problem

The multigateway binary supports TLS via `--pg-tls-cert-file` / `--pg-tls-key-file` (merged in multigres#659), but the operator wasn't creating certificates or passing these flags. This caused `psql sslmode=require` to fail with "server does not support SSL" for HA projects (MUL-326).

## Solution

When `certCommonName` is set on the `MultigresCluster` CR, the cell controller creates a cert-manager `Certificate` matching the non-HA project convention:

- Issuer: `supabase-issuer` (ClusterIssuer)
- Secret: `generated-certs`
- RSA 2048, 44640h duration (5 years)
- DNS SANs: `db.<ref>.<domain>` + `<ref>.<domain>`

The multigateway deployment then mounts the `generated-certs` secret and passes the TLS flags. This supports all PostgreSQL SSL modes including `verify-full`.

The platform worker (separate repo) will pass `certCommonName` in the `MultigresCluster` CR — a 1-line change on their side.

## Test plan

- [x] `TestBuildCertificate` — Certificate resource name, DNS SANs, owner reference, issuer, key params
- [x] `TestReconcileCertificate` — no-op when empty, creates via SSA, idempotent on repeated calls
- [x] `TestBuildMultiGatewayDeployment_TLS` — TLS enabled (volume, mount, args) and disabled (no artifacts)
- [x] Full test suite passes
- [ ] Deploy to staging with `certCommonName: db.<ref>.supabase.red` and verify `psql sslmode=require`


pkg/resource-handler/controller/cell/certificate.go
```diff
@@ -0,0 +1,103 @@
...
+		dnsNames = append(dnsNames, after)
+	}
+
+	cert := &unstructured.Unstructured{}
+	cert.SetGroupVersionKind(schema.GroupVersionKind{
+		Group:   "cert-manager.io",
+		Version: "v1",
+		Kind:    "Certificate",
+	})
+	cert.SetName(cn)
```
Verolop: this should use a single cluster-level owner/reconciler, or make certs explicitly per-cell with unique names.
The reason for this is that `certCommonName` is inherited by every cell, so multiple cells end up reconciling the same Certificate, and ownerRef keeps flipping between them.

pkg/resource-handler/controller/cell/certificate.go
```diff
@@ -0,0 +1,103 @@
...
+	return cert
+}
+
+// reconcileCertificate creates or updates the cert-manager Certificate for
+// multigateway TLS. This is a no-op if CertCommonName is empty.
+func (r *CellReconciler) reconcileCertificate(
+	ctx context.Context,
+	cell *multigresv1alpha1.Cell,
+) error {
+	if cell.Spec.CertCommonName == "" {
```
Verolop: can you please add cleanup for the previously-managed cert so TLS disable is deterministic?

pkg/resource-handler/controller/cell/certificate_test.go
```diff
@@ -0,0 +1,300 @@
...
+	// Register cert-manager Certificate as an unstructured type so the fake
+	// client can store and retrieve it. Without this the fake client falls
+	// back to its unstructured storage which happens to work today, but
+	// explicitly registering it makes the test contract clear.
+	scheme.AddKnownTypeWithName(certGVK, &unstructured.Unstructured{})
+	certListGVK := certGVK
+	certListGVK.Kind += "List"
+	scheme.AddKnownTypeWithName(certListGVK, &unstructured.UnstructuredList{})
+
+	t.Run("no-op when CertCommonName is empty", func(t *testing.T) {
```
Verolop: please add regression tests for:
- two cells with same CN in same namespace
- CN change
- CN unset cleanup

Verolop (changes requested):
nice work! 
Before merging, we need to have one clear owner for certificate lifecycle. Right now the config comes from one scope and the resource is managed from another, which creates ambiguity during updates and deletion paths.

pkg/resource-handler/controller/cell/certificate_test.go
haritabh17: Ack, added all.

pkg/resource-handler/controller/cell/certificate.go
haritabh17: Ack, added. When `certCommonName` is empty, `reconcileCertificate` now lists `cert-manager` Certificates in the namespace, filters by `ownerRef` UID matching this cluster, and deletes them.

pkg/resource-handler/controller/cell/certificate.go
haritabh17: Ack, moved buildCertificate and reconcileCertificate to the multigrescluster package. The Certificate's ownerRef now points to the MultigresCluster (not the Cell).

pkg/cluster-handler/controller/multigrescluster/certificate.go
```diff
@@ -0,0 +1,185 @@
...
+// Certificate. When CertCommonName is empty it deletes a previously-managed
+// Certificate (identified by the generated-certs secret convention) so that
+// disabling TLS is deterministic.
+func (r *MultigresClusterReconciler) reconcileCertificate(
+	ctx context.Context,
+	cluster *multigresv1alpha1.MultigresCluster,
+) error {
+	logger := log.FromContext(ctx)
+
+	if cluster.Spec.CertCommonName != "" {
```
Verolop: one last thing: on certCommonName changes, reconcile should also remove the cert from the previous CN. Right now, since the cert name is the CN, the change creates a second Certificate instead of updating the existing one. The old cert is still owned by the same MultigresCluster, and ownerRef GC won’t remove it unless the cluster itself is deleted, so it can stick around and still target secretName: generated-certs.

haritabh17: Ack, extracted `deleteOwnedCertificates(ctx, cluster, keepName)` which runs before the apply step on every reconcile.

Verolop (approved):
LGTM! ✨

PR #465, opened 2026-04-16
feat(operator): add cluster-wide multigateway replica-read endpoint

Verolop:
## Description

this pr adds a cluster-wide replica-read endpoint for multigateway while keeping the existing primary endpoint unchanged.

It enables the replica listener in multigateway pods by default and introduces a dedicated global Service so clients can explicitly target read-only traffic.

## Changes

- Enabled replica listener in cell multigateway deployments with `--pg-replica-port=5433`.
- Added container port `5433` named `postgres-replica` to multigateway pods.
- Added a new global Service builder for `<cluster>-multigateway-replica` as `ClusterIP`, using the same cluster-wide multigateway selector shape and exposing `5433 -> postgres-replica`.
- Updated global reconciliation to apply and own both Services:
  - `<cluster>-multigateway` for primary traffic on `5432`
  - `<cluster>-multigateway-replica` for replica traffic on `5433`

## Testing

Added coverage for the new replica service reconciliation error paths so failures are attributed to the replica endpoint instead of generic service patch ordering.

Updated integration expectations to assert the two-endpoint gateway contract explicitly: primary remains on `<cluster>-multigateway` and replica reads are served via `<cluster>-multigateway-replica`.

### DIFF
--- a/pkg/cluster-handler/controller/multigrescluster/builders_global.go
+++ b/pkg/cluster-handler/controller/multigrescluster/builders_global.go
@@ -15,6 +15,11 @@ import (
 	"github.com/multigres/multigres-operator/pkg/util/metadata"
 )
 
+const (
+	globalMultiGatewayReplicaPort       int32 = 5433
+	globalMultiGatewayReplicaTargetPort       = "postgres-replica"
+)
+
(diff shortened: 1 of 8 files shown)


GuptaManan100 (approved):
Love it! ❤️

PR #470, opened 2026-04-21
feat: add PostgresSuperuser to MultigresClusterSpec

soedirgo:
Passed to pgctld and multipooler as POSTGRES_USER env var, to allow initializing the db with a different bootstrap superuser name.

Right now if using a different bootstrap superuser name (e.g. `POSTGRES_USER=supabase_admin`) multipooler won't be able to connect as the `postgres` role (which is the default) would be missing.


pkg/resource-handler/controller/shard/containers.go
```diff
@@ -155,6 +155,7 @@ func buildPgctldContainer(
 			Name:  "PGDATA",
 			Value: PgDataPath,
 		},
+		pgUserEnvVar(shard.Spec.PostgresSuperuser),
```
Verolop, line 158: `postgres-exporter` still connects as postgres here (DATA_SOURCE_USER and /postgres), so a custom `spec.postgresSuperuser` would only be partially honoured. If we expose this as a first-class field, the exporter should use the same source of truth as the other operator-managed admin clients.

pkg/resource-handler/controller/shard/containers_test.go
```diff
@@ -105,6 +105,8 @@ func TestBuildMultiPoolerSidecar(t *testing.T) {
 				},
 				Env: []corev1.EnvVar{
 					{Name: "PGDATA", Value: PgDataPath},
+					{Name: "POSTGRES_USER", Value: "postgres"},
```
Verolop, line 108: This test covers the default postgres case, but it doesn't show that a custom `postgresSuperuser` actually makes it through to the generated env vars. Can you please add one focused test with something like `postgresSuperuser: supabase_admin` that verifies pgctld and multipooler pick it up?

pkg/cluster-handler/controller/multigrescluster/builders_global.go
```diff
@@ -299,7 +299,7 @@ func BuildMultiAdminWebDeployment(
 								},
 								{
 									Name:  "POSTGRES_USER",
-									Value: "postgres",
+									Value: postgresSuperuserOrDefault(cluster.Spec.PostgresSuperuser),
```
Verolop: Can you please also add a focused test for the non-default case here? You're in the right path here, but this still mostly locks in the default postgres value rather than proving that a custom `spec.postgresSuperuser` flows all the way into multiadmin-web.

Verolop (changes requested):
Thanks for putting this together! This makes sense for the operator. We already own the bootstrap/admin wiring for pgctld, multipooler, and multiadmin-web, so having a configurable bootstrap superuser feels like it fits here.

Please see my comments. Before merging it, we need the following:

- wire the configured superuser through every operator-managed Postgres client
- at least one focused test for a non-default value

soedirgo:
Thanks! Fixed the propagation to postgres-exporter and added tests for a custom `PostgresSuperuser` value

Verolop (approved):

PR #471, opened 2026-04-23
fix(observability): reconcile project-ref annotation updates on existing clusters

Verolop:
## Description

this pr fixes project-ref propagation for existing clusters when `multigres.com/project-ref` is added, changed, or removed after the cluster has already been created.

Create-time propagation already worked, but metadata-only updates on the parent resource did not fan out because the relevant controllers only reconciled on generation changes. As a result, child resources and runtime pods kept the old fallback value until the cluster was recreated.

## Changes

- Updated the `MultigresCluster` controller predicate to reconcile when `multigres.com/project-ref` changes, not just when generation changes.
- Updated the `TableGroup` controller predicate to reconcile when `multigres.com/project-ref` changes, not just when generation changes.
- Updated the `Cell` controller predicate to reconcile when `multigres.com/project-ref` changes, not just when generation changes.
- Kept the predicate scope narrow so unrelated metadata churn does not trigger extra reconciles.
## Testing

Verified locally that the affected controller packages still pass after the predicate change.

Also validated the failure mode in EKS before the fix: creating a cluster with `multigres.com/project-ref` set from the start correctly propagated the explicit value into runtime pods, while adding the annotation after creation did not update existing child resources or pods. This change addresses that update path by treating `multigres.com/project-ref` as reconciliation input for the relevant controllers.

### DIFF
--- a/pkg/resource-handler/controller/cell/cell_controller.go
+++ b/pkg/resource-handler/controller/cell/cell_controller.go
@@ -16,6 +16,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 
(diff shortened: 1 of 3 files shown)


pkg/resource-handler/controller/cell/cell_controller.go
```diff
@@ -393,9 +394,41 @@ func (r *CellReconciler) SetupWithManager(mgr ctrl.Manager, opts ...controller.O
...
 		Owns(&appsv1.Deployment{}).
 		Owns(&corev1.Service{}).
 		WithOptions(controllerOpts).
 		Complete(r)
 }
+
+// projectRefOrGenerationChangedPredicate requeues when desired gateway state can
+// change due to either spec updates or project-ref annotation updates.
+func projectRefOrGenerationChangedPredicate() predicate.Predicate {
+	return predicate.Funcs{
```
niconosenzo, line 409: Looks like we always get the same Predicate, should we extract this function into a shared package ?

niconosenzo (approved):
Apart from my comment above, it LGTM.

pkg/resource-handler/controller/cell/cell_controller.go
Verolop: I did consider splitting it, but decided to keep it local for now as part of the reconciliation contract (vs. business logic) and I wanted the reconcile trigger to stay visible at the `SetupWithManager` callsite for readability purposes for folks who need to interact with the Operator at this stage.

PR #472, opened 2026-04-23
fix: reduce default multigateway replicas per cell

Verolop:
## Description

this pr fixes the per-cell operator defaults so they match the intended HA baseline for multi-cell clusters.

The previous default created two pods per cell, which meant a three-cell cluster ran six steady-state gateway pods instead of the intended three. That doubled baseline gateway capacity and resource usage for clusters that did not explicitly override the template, even though rollouts can temporarily burst above the steady-state target when needed.

It also fixes the pool replica validation warning so a three-cell cluster with one postgres pod per cell is not warned as under-replicated for HA.

## Changes

- Changed the default multigateway replica count from `2` to `1` per cell.
- Updated the pool replica warning to evaluate total resolved pods across the pool's cells instead of only `replicasPerCell`.
- Reworded the warning to distinguish the `AT_LEAST_2` HA baseline from the extra local capacity needed for zero-downtime rolling upgrades within a single cell.
- Added validation coverage for the three-cell, one-replica-per-cell baseline.

## Testing

Brought up a basic HA cluster on EKS, scaled it down and verified the error message was gone for the previous case.

Ran the focused cell controller and resolver test packages to verify both sides of the change: multigateway reconciliation now creates deployments with the new default replica count, and resolver validation no longer warns for the intended three-cell HA baseline while still warning when total pool pods are below the threshold.

### DIFF
--- a/pkg/resolver/validation.go
+++ b/pkg/resolver/validation.go
@@ -453,21 +453,25 @@ func (r *Resolver) ValidateClusterLogic(
 				}
 
 				// ------------------------------------------------------------------
-				// 3. Quorum Warning for pools with insufficient replicas
+				// 3. Quorum Warning for pools with insufficient total replicas
 				// ------------------------------------------------------------------
 				for poolName, pool := range pools {
 					replicas := int32(3) // default
 					if pool.ReplicasPerCell != nil {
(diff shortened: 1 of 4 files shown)


GuptaManan100 (approved):
💖

PR #473, opened 2026-04-24
fix: populate QuorumType and RequiredCount in DurabilityPolicy

soedirgo:
Extract buildDurabilityPolicy helper that sets QuorumType and RequiredCount based on the policy name. Without these fields, multiorch's analyzers fail with "unsupported quorum type" because the proto3 zero value (QUORUM_TYPE_UNKNOWN) is not handled.

Without this fix, when using the `supabase/postgres` image, we get this error from the MultiOrch:

```
{"time":"2026-04-24T08:02:11.875452119Z","level":"ERROR","msg":"analyzer error","analyzer":"ShardNeedsInitialization","shard":{"Database":"postgres","TableGroup":"default","Shard":"0-inf"},"error":"failed to parse bootstrap durability policy: unsupported quorum type: QUORUM_TYPE_UNKNOWN","trace_id":"4c2893eb891408d024e685ba0fb92089","span_id":"f56597b2c0ea8fdf"}
```

We didn't get this error when using `multigres/pgctld` image; according to Claude, this is because:

> The operator writes an incomplete `DurabilityPolicy` protobuf into etcd — it sets `PolicyName: "AT_LEAST_2"` but leaves `QuorumType` and `RequiredCount` at their proto3 zero values (`QUORUM_TYPE_UNKNOWN`, `0`).

> Upstream multigres' `NewPolicyFromProto` routes purely on `QuorumType`, so it rejects the stored record with `unsupported quorum type: QUORUM_TYPE_UNKNOWN`. That error bubbles up from multiorch's `ShardNeedsInitialization` analyzer, which then never emits the problem that would trigger `shard_init_action`, so no primary is ever elected.

> The bug is image-agnostic, but with `multigres/pgctld` the cohort forms fast enough that the analyzer short-circuits before reaching the broken parse; with `supabase/postgres` the slower bootstrap lets the analyzer hit the faulty proto on every tick and bootstrap stalls forever.

### DIFF
--- a/pkg/data-handler/topo/database.go
+++ b/pkg/data-handler/topo/database.go
@@ -151,11 +151,26 @@ func GetBackupLocation(shard *multigresv1alpha1.Shard) *clustermetadatapb.Backup
 // GetDurabilityPolicy extracts the durability policy from the shard config.
 // Falls back to "AT_LEAST_2" if not set (the default materialized by the webhook resolver).
 func GetDurabilityPolicy(shard *multigresv1alpha1.Shard) *clustermetadatapb.DurabilityPolicy {
-	name := "AT_LEAST_2"
-	if shard.Spec.DurabilityPolicy != "" {
-		name = shard.Spec.DurabilityPolicy
+	return buildDurabilityPolicy(shard.Spec.DurabilityPolicy)
+}
+
+// buildDurabilityPolicy constructs a fully populated DurabilityPolicy proto from
+// a policy name. Upstream multigres' consensus code routes on QuorumType, so
+// leaving it at the proto3 zero value (QUORUM_TYPE_UNKNOWN) would cause
+// "unsupported quorum type" failures in multiorch's analyzers. Falls back to
+// "AT_LEAST_2" when name is empty.
+func buildDurabilityPolicy(name string) *clustermetadatapb.DurabilityPolicy {
+	if name == "" {
+		name = "AT_LEAST_2"
 	}
-	return &clustermetadatapb.DurabilityPolicy{
-		PolicyName: name,
+	policy := &clustermetadatapb.DurabilityPolicy{PolicyName: name}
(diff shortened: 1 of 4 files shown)


pkg/data-handler/topo/database.go
```diff
@@ -151,11 +151,26 @@ func GetBackupLocation(shard *multigresv1alpha1.Shard) *clustermetadatapb.Backup
...
-	return &clustermetadatapb.DurabilityPolicy{
-		PolicyName: name,
+	policy := &clustermetadatapb.DurabilityPolicy{PolicyName: name}
+	switch name {
+	case "AT_LEAST_2":
+		policy.QuorumType = clustermetadatapb.QuorumType_QUORUM_TYPE_AT_LEAST_N
+		policy.RequiredCount = 2
+	case "MULTI_CELL_AT_LEAST_2":
+		policy.QuorumType = clustermetadatapb.QuorumType_QUORUM_TYPE_MULTI_CELL_AT_LEAST_N
+		policy.RequiredCount = 2
```
niconosenzo, line 187: It makes sense. Just one thought, do you think we can use the [upstream durability policy parser](https://github.com/multigres/multigres/blob/e76358092d4dd0233791010ff6afdf94c40b2105/go/common/consensus/durability.go#L28) for consistency ?

rafael: I think for now, we can keep it like this. The one in upstream I added super recently and it was kind of an internal helper. 

So would like to avoid a dependency between the two projects on this particular method. 

One nit: is that we should return error if the policy doesn't match exactly. This code will still default to UNKNOWN and I think we don't want that

soedirgo: Updated to return an error on unexpected policy name 👍

Verolop (approved):

PR #476, opened 2026-04-29
feat: add zone ID support to MultigresCluster cell definitions

niconosenzo:
## Description

Cells are now identified by AZ ID (e.g. use1-az1) instead of zone name  (e.g. us-east-1a). AZ IDs are consistent across AWS accounts, which makes them a better fit for cross-account topology constraints. Region remains supported as an alternative for coarser-grained placement.

## Changes

API changes:
- Remove Zone type and zone field from CellConfig (MultigresCluster CRD)
  and CellSpec (Cell CRD)
- Add ZoneID type to common_types.go
- Add zoneId field to CellConfig and CellSpec
- CEL validation: zoneId and region are mutually exclusive; at least one
  of zoneId or region must be specified
- Regenerate CRD manifests for multigres.com_cells and
  multigres.com_multigresclusters

Controller changes:
- BuildCell propagates zoneId from CellConfig to CellSpec
- buildCellTopologyLabels uses topology.k8s.aws/zone-id for zone ID cells
- buildCellNodeSelector uses topology.k8s.aws/zone-id for zone ID cells,
  matching the node label set by the AWS node labeler
- validateCellTopology checks topology.k8s.aws/zone-id node labels
- Centralize topology.k8s.aws/zone-id as metadata.NodeLabelZoneID constant

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -382,7 +382,7 @@ kind-up: ## Create a kind cluster for local development
 	@echo "==> Cluster ready. Use: export KUBECONFIG=$(KIND_KUBECONFIG)"
 
 .PHONY: kind-up-topology
-kind-up-topology: ## Create a multi-node kind cluster with topology zone labels
+kind-up-topology: ## Create a multi-node kind cluster with topology zone ID labels
 	@command -v $(KIND) >/dev/null 2>&1 || { \
 		echo "ERROR: kind is not installed."; \
 		echo "Install it from: https://kind.sigs.k8s.io/docs/user/quick-start/"; \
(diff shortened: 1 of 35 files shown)


.claude/settings.local.json
```diff
@@ -0,0 +1,8 @@
+{
```
Verolop: can we drop this from the pr?

Verolop (approved):
Thanks for reworking this @niconosenzo !
Left a small cleanup comment.

Also, not blocking, but I noticed a few existing samples/docs still mention the old zone field, like `config/samples/topology.yaml` and `docs/development/cell-topology.md`. Can you please update those as well so users don’t copy examples that fail validation? Either here or a follow up pr.

Verolop (approved):

PR #479, opened 2026-05-05
ci: auto-sync multigres proto dependency via repository dispatch

niconosenzo:
Adds workflow that listens for proto-updated repository_dispatch events from multigres, validates the incoming SHA is reachable from `multigres/main` (rejects forks, unmerged PR heads, and arbitrary branches), bumps the multigres dependency via `go get` + `go mod tidy`, and opens a PR with the resulting `go.mod` and `go.sum` changes.

### DIFF
--- a/.github/workflows/sync-multigres-proto.yml
+++ b/.github/workflows/sync-multigres-proto.yml
@@ -1,0 +1,70 @@
+name: Sync Multigres Proto
+on:
+  repository_dispatch:
+    types: [proto-updated]
+
+permissions: {}
+
+concurrency:
+  group: sync-proto
+  cancel-in-progress: false
+
+jobs:
+  update-dependency:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Generate GitHub App token for creating a PR
+        id: app-token
+        uses: actions/create-github-app-token@f8d387b68d61c58ab83c6c016672934102569859 # v3.0.0
+        with:
+          app-id: ${{ vars.MULTIGRES_BOT_APP_ID }}
+          private-key: ${{ secrets.MULTIGRES_BOT_APP_PRIVATE_KEY }}
+          owner: multigres
+          repositories: multigres,multigres-operator
+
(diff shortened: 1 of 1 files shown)


.github/workflows/build-branch.yaml
```diff
@@ -0,0 +1,27 @@
...
+      - main
+    paths-ignore:
+      - '**.md'
+      - 'docs/**'
+      - 'LICENSE'
+      - 'CODEOWNERS'
+      - '.gitignore'
+      - '.editorconfig'
+      - '.github/**'
+
```
Verolop: Can we request explicit App token permissions here instead of inheriting the full installation permissions? Narrowing it to the required repo read + contents/PR write permissions would reduce blast radius if it’s ever misused or leaked.

.github/workflows/sync-multigres-proto.yml
```diff
@@ -0,0 +1,70 @@
...
+permissions: {}
+
+concurrency:
+  group: sync-proto
+  cancel-in-progress: false
+
+jobs:
+  update-dependency:
+    runs-on: ubuntu-latest
+    steps:
```
Verolop, line 15: I think this also needs security-events: write because the reusable build workflow uploads grype.sarif with github/codeql-action/upload-sarif. Called workflows can’t get broader permissions than the caller grants, and I see main.yaml already includes this, so branch builds may fail at the sarif upload without it.

Verolop (commented):
Please see my comments around permissions and workflow scope. Other than that, the important chunks lgtm.

.github/workflows/build-branch.yaml
niconosenzo: Dropped this one, it was unintended.

.github/workflows/sync-multigres-proto.yml
niconosenzo: Removed the unwanted workflow file, permissions should be ok now.

Verolop (approved):

PR #480, opened 2026-05-05
feat: use `POSTGRES_INITDB_EXTRA_CONF` for custom pg configs

soedirgo:
Update supplying of custom postgresql.conf to use `POSTGRES_INITDB_EXTRA_CONF` instead of `--postgres-config-template`. This is consistent with the purpose of the env - overriding the config template is a more complicated ordeal because pgctld expects certain params to be in it and it's parsed as a Go template file.

Follow up to https://github.com/multigres/multigres/pull/949

### DIFF
--- a/api/v1alpha1/common_types.go
+++ b/api/v1alpha1/common_types.go
@@ -263,7 +263,8 @@ type CellName string
 // +kubebuilder:validation:MaxLength=512
 type InitdbArgs string
 
-// PostgresConfigRef references a ConfigMap containing custom postgresql.conf content.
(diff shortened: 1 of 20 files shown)


Verolop (commented):
looks good, but can you please bump the default pgctld image, and any related default Multigres images if needed, to a build containing `multigres/multigres#949` / commit `54ab7639` or newer ? so the changes on the Multigres side are compatible with this one too

soedirgo:
Done 👍

Verolop (approved):

PR #481, opened 2026-05-05
fix: correct AT_LEAST_2 pool replica semantics

Verolop:
## Description

this pr fixes the operator’s previous interpretation of `AT_LEAST_2` pool sizing. Follow-up to #472 .

`replicasPerCell` is applied per selected cell, but the operator was treating `3` as the durability baseline. That was incorrect for Multigres: `AT_LEAST_2` requires 2 total poolers, while `MULTI_CELL_AT_LEAST_2` requires poolers in 2 distinct cells. Three total poolers is HA operating guidance so the policy can remain achievable while one pooler is unavailable, not the policy minimum.

## Changes

- Default pools to `replicasPerCell: 1`, matching the field’s per-cell meaning.
- Update durability validation to check the resolved pool shape against the actual policy semantics:
  - `AT_LEAST_2`: at least 2 total poolers.
  - `MULTI_CELL_AT_LEAST_2`: at least 2 selected cells with poolers.
- Keep stored webhook defaults dynamic by using cell context without materializing cell lists into the stored spec.
- Share durability policy inheritance logic between tablegroup resolution and validation.
- Update API comments, generated CRD descriptions, samples, and docs so 3 total poolers is documented as HA guidance rather than the policy minimum.

## Tests

Test coverage was updated around the resolved pool shapes that matter for these policies: single-cell pools below and at the `AT_LEAST_2` minimum, multi-cell pools with one replica per cell, omitted `replicasPerCell` defaults, and `MULTI_CELL_AT_LEAST_2` cases with one, two, and three cells.

### DIFF
--- a/README.md
+++ b/README.md
@@ -252,7 +252,7 @@ Multigres uses a configurable **durability policy** to control synchronous repli
 | **2** | 1 primary + 1 standby | **Downtime during upgrades.** Draining the standby leaves zero synchronous standbys, violating `AT_LEAST_2`. Upstream multigres rejects the `UpdateSynchronousStandbyList REMOVE` because it would empty the synchronous standby list. |
 | **3** (recommended) | 1 primary + 2 standbys | **Zero-downtime upgrades.** One standby can be drained while the other maintains quorum. |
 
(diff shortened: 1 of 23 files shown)


Verolop:
@GuptaManan100 requested your review since you reviewed the 1st part of this small refactor ❤️ https://github.com/multigres/multigres-operator/pull/472

GuptaManan100 (approved):

PR #493, opened 2026-05-13
feat(topology): support shared local topology registration

Verolop:
## Description

this pr adds operator-side support for shared topology layouts where Multigres components use a shared global topology server and each Cell points to its own local topology root. The operator now registers global Cell records with the resolved local topology endpoint/root, while keeping cluster level Database records in global topology.

## Main Changes

- Register Cell metadata using the cell’s local topology config when available, falling back to global topology for existing/default behaviour.
- Make Cell registration idempotent by updating stale existing Cell records during reconciliation.
- Resolve cell topology templates/defaults during MultigresCluster topology reconciliation before writing Cell records.
- Add tests for shared topo registration, stale Cell updates, fallback behaviour, and global Database registration.

## Testing

Tests cover direct Cell registration, repeated reconciliation against existing topology records, fallback behaviour when no local topology is configured, and the cluster-level reconciliation path that registers Cells and Databases together.


rafael (approved):
LGTM. This makes sense

PR #496, opened 2026-05-15
feat(gc): add garbage collector for orphaned multigres resources

niconosenzo:
- Introduce pkg/gc with a pluggable Cleankeeper interface so additional resource kinds can be added with minimal effort.
- Add pkg/gc/pvc Cleankeeper that deletes PVCs labeled `app.kubernetes.io/managed-by=multigres-operator` and `multigres.com/orphan-since=<YYYY-MM-DD>` once past the configured retention window (default 30d), with a resourceVersion precondition to avoid racing with operator re-adoption.
- Add `cmd/multigres-gc`, a standalone one-shot binary intended to be invoked from a cronjob (to be added in a follow-up PR).
- Build multigres-gc alongside manager in the Dockerfile and expose a build-multigres-gc Makefile target.
- Add multigres.com/orphan-since label and OrphanDateFormat constants in `pkg/util/metadata`.

### DIFF
--- a/cmd/multigres-gc/main.go
+++ b/cmd/multigres-gc/main.go
@@ -1,0 +1,136 @@
+// The multigres-gc command runs a one-shot clean of orphaned multigres resources.
+//
+// It is intended to be invoked from a k8s cronjob, separate from the
+// operator's reconcile loop. New resource types are added by implementing
+// gc.Cleankeeper and registering them in registerCleankeepers().
+package main
+
+import (
+	"flag"
+	"fmt"
+	"os"
+	"time"
+
+	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
+	_ "k8s.io/client-go/plugin/pkg/client/auth"
+
+	"github.com/go-logr/logr"
+	"k8s.io/apimachinery/pkg/runtime"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/log/zap"
+
+	"github.com/multigres/multigres-operator/pkg/gc"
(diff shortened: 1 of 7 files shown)


pkg/gc/pvc/pvc.go
```diff
@@ -0,0 +1,153 @@
...
+
+	raw, ok := pvc.Labels[metadata.LabelOrphan]
+	if !ok {
+		// Selector guarantees presence; treat as malformed if absent.
+		log.Info("orphan label missing despite selector match")
+		res.Malformed++
+		return
+	}
+
+	orphanDate, err := time.Parse(metadata.OrphanDateFormat, raw)
```
Verolop: nit: since this parses `YYYY-MM-DD` as midnight UTC, PVCs orphaned later that day could become eligible almost 24h earlier than intended. Can you either store a full timestamp, or make sure to only delete after the labeled day plus the retention window has already gone by? It's an edge case, but at scale it matters.

cmd/multigres-gc/main.go
```diff
@@ -0,0 +1,125 @@
...
+}
+
+func main() {
+	var (
+		retention time.Duration
+		namespace string
+		dryRun    bool
+	)
+
+	flag.DurationVar(&retention, "retention", 30*24*time.Hour,
```
Verolop, line 50: can negative retention values be rejected before running the cleaners? e.g. `-retention=-1h` would make even newly marked PVCs eligible right away. We should probably also decide whether `0` is allowed as “delete immediately” or should be rejected too 🤔

Verolop (commented):
Looking good! I left two small comments around retention details.

pkg/gc/pvc/pvc.go
niconosenzo: Good catch, lets use a full timestamp.

cmd/multigres-gc/main.go
niconosenzo: 👍, Yeah I think that for immediate deletions, we can cover them by setting up retention to 0, it should be explicit enough to say "keep them around for 0 hours once the object has been deleted"

Verolop (approved):
lgtm!

PR #502, opened 2026-05-19
feat(shard): orphan PVCs instead of deleting them on small shards

niconosenzo:
Until now the shard controller relied on k8s cascade-GC (via a controller ownerRef on each PVC) and directly triggered delete calls to clean up data volumes when a pod was drained, scaled down, or the whole Shard was removed. 

## Changes
Hand the cleanup to multigres-gc instead, so PVCs stick around long enough for an operator to roll back a shard deletion:

  - Add pkg/util/pvc.MarkOrphan, which strips a given ownerRef from a PVC and stamps it with multigres.com/orphan-since=<UTC now>. Idempotent: an existing orphan-since label is preserved so the retention window starts at the first orphan event, not at every reconcile.

  - Gate Shard deletion on a new finalizer (multigres.com/shard-pvc-orphan). Without it, cascade-GC fires the instant the Shard's DeletionTimestamp is set and we have no chance to detach the PVCs first.

  - Rename deletePodPVC -> cleanupPodPVC and replace the in-line delete with MarkOrphan for drained pods and scale-down events.

  - Replace handleDeletion's reliance on ownerRef cascade with a new cleanupShardPVCs step that enumerates Shard-scoped PVCs (via the cluster/database/tablegroup/shard label set) and orphans those whose effective PVCDeletionPolicy resolves to Delete.

Large shards skip the orphan path. Once shard.Spec.Replicas exceeds pvcOrphanReplicasThreshold (3) the operator deletes the PVC in-line at both cleanup sites, for those workloads keeping every per-replica volume around for the retention window would be too expensive. nil Replicas (Shard CRs created before the field existed) default to the small-shard path.

### DIFF
--- a/pkg/resource-handler/controller/shard/reconcile_deletion.go
+++ b/pkg/resource-handler/controller/shard/reconcile_deletion.go
@@ -3,6 +3,8 @@ package shard
 import (
 	"context"
 	"fmt"
+	"slices"
+	"strings"
 	"time"
 
 	"github.com/multigres/multigres/go/common/topoclient"
@@ -17,6 +19,7 @@ import (
 	multigresv1alpha1 "github.com/multigres/multigres-operator/api/v1alpha1"
 	"github.com/multigres/multigres-operator/pkg/data-handler/drain"
 	"github.com/multigres/multigres-operator/pkg/util/metadata"
+	pvcutil "github.com/multigres/multigres-operator/pkg/util/pvc"
 	"github.com/multigres/multigres-operator/pkg/util/status"
 )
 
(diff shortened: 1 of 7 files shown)


Verolop:
the important bits are covered, so that's great. I’d add two lifecycle rules so the GC handoff is super clear:

1st one: `orphan-since` should only be on a PVC while it’s actually orphaned. Since PVC names are deterministic, the same shard/pool/cell/index can become desired again later. If that happens, reconcile should clear the label before starting a pod on that PVC, otherwise the GC can still delete storage we’ve put back in use. e.g. a scale down marks a deterministic PVC orphan, then the scaling back up reuses the same PVC before GC runs. If the orphan label is still there, the GC can delete storage that has become live again. 

and then, I’d be explicit that the PVC only gets handed to GC after the pod using it is gone. The GC can own the delayed delete, but the operator should still decide when a live member has become orphaned storage. Your proposal already does this implicitly, but e.g. in a scenario where a pod deletion gets stuck because the node is unhealthy, if we mark the PVC orphan as soon as deletion starts, GC may eventually delete the backing volume while Kubernetes still has a pod object referencing it, so to be safe, delete the pod, wait until the pod is actually gone, and then mark the PVC for GC.

wdyt?

niconosenzo:
Good catch, reg 2, we should be able to cover it by slightly modifying the order of events: delete pods first, requeue while any still exists, run `cleanupShardPVCs` only after the pod list is empty, then drop the finalizer.

Verolop (approved):
yay!

PR #508, opened 2026-05-21
feat: use password files for pool postgres credentials

Verolop:
## Description

this PR add the operator-side companion change for upstream Multigres `POSTGRES_PASSWORD_FILE` support. Pool pods now mount the existing per-shard postgres password Secret as a file and pass that file path to pgctld and multipooler, while postgres-exporter continues using its existing SecretKeyRef-based password wiring.

## Main Changes

- Mount the per-shard postgres password Secret into pool pods as a read-only file.
- Switch pgctld and multipooler from `POSTGRES_PASSWORD` to `POSTGRES_PASSWORD_FILE`.
- Keep postgres-exporter on `DATA_SOURCE_PASS`.
- Include the new password volume and mount details in the pool pod spec hash so existing pods roll.
- Bump Multigres runtime image defaults to the upstream SHA containing `POSTGRES_PASSWORD_FILE` support.
- Document the Secret volume file mode choice for non-root pool containers.

## Testing

New tests cover the full pool pod contract for the credential migration: the password Secret is present as a volume, pgctld and multipooler mount it read-only, both containers receive `POSTGRES_PASSWORD_FILE`, and neither receives the legacy `POSTGRES_PASSWORD`. They also verify postgres-exporter remains on its existing `DATA_SOURCE_PASS` SecretKeyRef path.

The spec-hash tests simulate old and drifted pod specs to ensure the new env var, Secret volume, volume mount, and mount read-only bit all contribute to the hash. That gives us coverage that existing pods will roll onto the file-based password wiring.

### DIFF
--- a/api/v1alpha1/image_defaults.go
+++ b/api/v1alpha1/image_defaults.go
@@ -5,25 +5,25 @@ package v1alpha1
 const (
 	// DefaultPostgresImage is the default container image for PostgreSQL instances.
 	// Uses the pgctld image which bundles PostgreSQL, pgctld, and pgbackrest.
-	DefaultPostgresImage = "ghcr.io/multigres/pgctld:sha-9d84da5"
+	DefaultPostgresImage = "ghcr.io/multigres/pgctld:sha-ba3ed8e"
 
 	// DefaultEtcdImage is the default container image for the managed Etcd cluster.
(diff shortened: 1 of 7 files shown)


mkindahl (approved):
LGTM

PR #511, opened 2026-05-22
feat: support external postgres password secrets

Verolop:
## Description

this PR lets the operator consume a pre-created Kubernetes Secret for the Postgres superuser password. Deployments that already generate or manage the final Postgres credentials can pass that Secret into Multigres at cluster creation time, instead of relying on the operator-generated default password and reconciling credentials later.

The change keeps the credential flow aligned with Multigres’ file-based password handling: components read the password from mounted Secret files, and the operator treats externally provided Secrets as inputs, not as resources it owns.

This PR also updates the default `pg_hba.conf` template so standby pods can establish SCRAM-authenticated replication connections to the primary from pod IPs. This unblocks pod-to-pod replication without reintroducing `trust` or `peer` authentication for TCP replication. Follow-up hardening for narrowing the replication source boundary is tracked separately.

## Main Changes

- Added `spec.postgresPasswordSecretRef` to `MultigresCluster` and propagate it through `TableGroup` and `Shard`.
- Mount the referenced Secret key at the stable Postgres password file path used by pgctld and multipooler.
- Updated postgres-exporter to use `DATA_SOURCE_PASS_FILE` instead of `DATA_SOURCE_PASS`.
- Validate externally referenced Secrets without owning, mutating, deleting, or watching them.
- Preserve the legacy generated per-shard password Secret path when no external Secret reference is set.
- Updated replication HBA rules from loopback-only to pod-to-pod TCP sources while keeping SCRAM mandatory.
- Regenerated CRDs and deepcopy code for the new API field.

## Testing

tests cover Secret ref propagation from `MultigresCluster` to pool pods, default and custom Secret keys, file-based password wiring for pgctld, multipooler, and postgres-exporter, and validation of externally managed Secrets without taking ownership.

It also covers the legacy generated-Secret path and verifies the HBA template keeps replication SCRAM-authenticated while allowing standby pods to connect from pod IPs.


api/v1alpha1/common_types.go
```diff
@@ -217,6 +217,26 @@ type PgBackRestTLSConfig struct {
...
+	Name string `json:"name"`
+
+	// Key is the Secret data key containing the password.
+	// Defaults to "password".
+	// +optional
+	// +kubebuilder:default="password"
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=253
+	// +kubebuilder:validation:Pattern=`^[-._a-zA-Z0-9]+$`
+	Key string `json:"key,omitempty"`
```
mkindahl: It might be a good idea to not allow a default and not allow it to be empty, i.e., something like this:

```suggestion
    // Key is the Secret data key containing the password.
    // +kubebuilder:validation:MinLength=1
    // +kubebuilder:validation:MaxLength=253
    // +kubebuilder:validation:Pattern=`^[-._a-zA-Z0-9]+$`
    Key string `json:"key"`
```

Also curious why it is restricted to the characters above, but we can loosen that later if we need to, so not important for this PR.

api/v1alpha1/multigrescluster_types.go
```diff
@@ -143,6 +143,12 @@ type MultigresClusterSpec struct {
 	// +kubebuilder:validation:MaxLength=63
 	PostgresSuperuser string `json:"postgresSuperuser,omitempty"`
 
+	// PostgresPasswordSecretRef references a Secret containing the final
+	// Postgres superuser password. When unset, the operator creates a per-shard
+	// Secret with the legacy default password.
+	// +optional
+	PostgresPasswordSecretRef *PostgresPasswordSecretRef `json:"postgresPasswordSecretRef,omitempty"`
```
mkindahl: Not sure how this works, but if we can enforce a password we reduce the risk of mistakes. It could be necessary to have no secret for pragmatic, and maybe transient, reasons though.

Update: Just saw the YAML below that it can be unset if not specified, but will in that case be filled in. Is that the case above as well?

config/crd/bases/multigres.com_multigresclusters.yaml
```diff
@@ -10519,6 +10519,30 @@ spec:
...
                     type: string
                 type: object
+              postgresPasswordSecretRef:
+                description: |-
+                  PostgresPasswordSecretRef references a Secret containing the final
+                  Postgres superuser password. When unset, the operator creates a per-shard
+                  Secret with the legacy default password.
+                properties:
+                  key:
+                    default: password
```
mkindahl: Do we want to support a default? If we can avoid one, it will probably reduce the risk mistakes. Right now it looks like you can accidentally omit the password and get the default.

pkg/resource-handler/controller/shard/templates/pg_hba_template.conf
```diff
@@ -20,13 +20,12 @@
...
-# instance shares the same postgres password, so the same pgpass file works
-# for both pgbackrest and replication.
+# Replication connections. Standbys authenticate via SCRAM using a passfile
+# pointed to by primary_conninfo (passfile=...). TCP rules are required because
+# standbys run in separate pods and connect from pod IPs, not loopback.
 local   replication     all                                     scram-sha-256
-host    replication     all             127.0.0.1/32            scram-sha-256
-host    replication     all             ::1/128                 scram-sha-256
+host    replication     all             0.0.0.0/0               scram-sha-256
+host    replication     all             ::0/0                   scram-sha-256
```
mkindahl, line 28: Looks good.

api/v1alpha1/shard_types.go
```diff
@@ -192,6 +192,11 @@ type ShardSpec struct {
 	// +kubebuilder:validation:MaxLength=63
 	PostgresSuperuser string `json:"postgresSuperuser,omitempty"`
 
+	// PostgresPasswordSecretRef is the resolved Postgres superuser password
+	// Secret.
+	// +optional
+	PostgresPasswordSecretRef *PostgresPasswordSecretRef `json:"postgresPasswordSecretRef,omitempty"`
```
mkindahl: Do we want it to be optional? Not sure what happens if we do not allow it.

```suggestion
	PostgresPasswordSecretRef *PostgresPasswordSecretRef `json:"postgresPasswordSecretRef"`
```

mkindahl (commented):
Overall it looks good, but it looks like it is possible to start a cluster with a default password, which seems avoidable. I am not sure what the consequences are if we enforce providing a password, but if we can do that, it would avoid some risk.

mkindahl (approved):
LGTM

PR #516, opened 2026-05-25
fix: update release process to generate crds and rbacs separately

kangmingtay:
* Generate separate manifests for CRDs and RBAC


Verolop (approved):
lgtm!

PR #517, opened 2026-05-25
fix: use correct sha for image versions

kangmingtay:
* The other sha specified does not point to a valid commit - it should use https://github.com/multigres/multigres/commit/5b9a7c6e32fb5c8224a51622fcb7ccaf0e74ced7 instead which corresponds to https://github.com/multigres/multigres/pkgs/container/multigres/881433391?tag=sha-5b9a7c6

### DIFF
--- a/api/v1alpha1/image_defaults.go
+++ b/api/v1alpha1/image_defaults.go
@@ -5,25 +5,25 @@ package v1alpha1
 const (
 	// DefaultPostgresImage is the default container image for PostgreSQL instances.
 	// Uses the pgctld image which bundles PostgreSQL, pgctld, and pgbackrest.
-	DefaultPostgresImage = "ghcr.io/multigres/pgctld:sha-ba3ed8e"
+	DefaultPostgresImage = "ghcr.io/multigres/pgctld:sha-5b9a7c6"
 
 	// DefaultEtcdImage is the default container image for the managed Etcd cluster.
 	DefaultEtcdImage = "gcr.io/etcd-development/etcd:v3.6.7"
 
 	// DefaultMultiAdminImage is the default container image for the MultiAdmin component.
-	DefaultMultiAdminImage = "ghcr.io/multigres/multigres:sha-ba3ed8e"
+	DefaultMultiAdminImage = "ghcr.io/multigres/multigres:sha-5b9a7c6"
 
 	// DefaultMultiAdminWebImage is the default container image for the MultiAdminWeb component.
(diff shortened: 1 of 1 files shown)


Verolop (changes requested):
thanks for jumping on this, @kangmingtay !  Though this PR is moving in the right direction, one extra piece is needed for the image pull issue. 
The cluster isn’t pulling runtime images based on go.mod; it’s pulling the compiled defaults from [api/v1alpha1/image_defaults.go](https://github.com/multigres/multigres-operator/blob/main/api/v1alpha1/image_defaults.go). Those still point at `sha-ba3ed8e` for the runtime images:

- DefaultPostgresImage: https://github.com/multigres/multigres-operator/blob/10e58acff6350cd6cccac2ed81477104ad699b74/api/v1alpha1/image_defaults.go#L8
- DefaultMultiAdminImage: https://github.com/multigres/multigres-operator/blob/10e58acff6350cd6cccac2ed81477104ad699b74/api/v1alpha1/image_defaults.go#L14
- DefaultMultiOrchImage / DefaultMultiPoolerImage / DefaultMultiGatewayImage: https://github.com/multigres/multigres-operator/blob/10e58acff6350cd6cccac2ed81477104ad699b74/api/v1alpha1/image_defaults.go#L20-L26

Those defaults are then copied into the resolved cluster spec when users don’t provide explicit images:

- Postgres / MultiAdmin: https://github.com/multigres/multigres-operator/blob/10e58acff6350cd6cccac2ed81477104ad699b74/pkg/resolver/cluster.go#L21-L26
- MultiOrch / MultiPooler / MultiGateway: https://github.com/multigres/multigres-operator/blob/10e58acff6350cd6cccac2ed81477104ad699b74/pkg/resolver/cluster.go#L30-L37

What happened is that `ba3ed8e` was an intermediate head of [multigres/multigres#1005](https://github.com/multigres/multigres/pull/1005)  while that PR was still being force-pushed. The PR eventually merged as `5b9a7c6`, and later the sync workflow advanced the operator Go dependency to d2c61f2. Since Multigres only publishes sha-* images from main pushes, sha-ba3ed8e never got images in GHCR even though the Git commit exists.

So I think the best fix is to keep this PR and add the image default bump here too:

- ghcr.io/multigres/multigres:sha-d2c61f2
- ghcr.io/multigres/pgctld:sha-d2c61f2

I checked that both of those tags exist. 

Longer term, we should add a CI guard or extend the sync workflow so any default image tag in image_defaults.go is validated with docker manifest inspect before merge.

Verolop (approved):
lgtm!

PR #522, opened 2026-05-27
fix(release): trigger release workflow on RC and prerelease tags

niconosenzo:
Add a second tag pattern `v[0-9]+.[0-9]+.[0-9]+-*` so any semver prerelease suffix (`-rc.N`, `-beta.N`, `-alpha.N`, etc.) triggers the same stable lane pipeline. GoReleaser's `prerelease: auto` setting then flags the resulting GitHub Release as a pre-release automatically, and the image is tagged with the full prerelease ref. The `latest` and `current` rolling tags remain untouched because they are only updated by the current lane on `main` pushes.

Also tighten `RELEASE.md` to spell out the accepted prerelease suffixes and clarify the rolling-tag behaviour.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -4,6 +4,7 @@ on:
   push:
     tags:
       - 'v[0-9]+.[0-9]+.[0-9]+'
+      - 'v[0-9]+.[0-9]+.[0-9]+-*'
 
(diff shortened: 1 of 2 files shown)


Verolop (approved):
thank you! lgtm

PR #524, opened 2026-05-29
fix(e2e-test): look for postgres container within initContainers

niconosenzo:
e2e tests have been failing for a while because of several reasons, this is an effort to fix all (or most) of them:

1- Native sidecar lookup
Commit 7330bc6 moved pgctld (the "postgres" container) from `Spec.Containers` to `Spec.InitContainers` as a native sidecar. The E2E helpers only scanned Spec.Containers / Status.ContainerStatuses, so WaitForPod and WaitForQueryServing never found "postgres" and timed out. Both helpers now also scan init containers.

2- Zone-scoped scheduling
Cells carry zoneId/region, which the operator turns into pod nodeSelectors (`topology.k8s.aws/zone-id`, `topology.kubernetes.io/region`). Kind nodes have no such labels, so zone-scoped pods stayed Pending. Label all Kind nodes on cluster setup and normalize fixture/sample zones to a single value so a single-node Kind cluster can schedule them. 

3- Numeric runAsUser for pool containers
Commits 08cedbc / 8d48d47 made container RunAsUser conditional on fsGroup. With no fsGroup, containers got `runAsNonRoot=true` and no numeric uid, kubelet cannot verify the pgctld/multigres images (USER postgres) are non-root and refused to start them (CreateContainerConfigError). Pin fsGroup=999 on CI pools.

4- Cache-sync detection
startManager relied on WaitForCacheSync returning false on a cancelled context, but with an empty scheme it returns true immediately (no informers to wait on), so the "cache sync fails" case never fired `t.Fatal`. Also treat a cancelled/expired context as a sync failure.

5- Durability vs replica count
The default durability policy is AT_LEAST_2, which needs >= 2 poolers for multiorch to elect a primary. The minimal sample's injected default pool has 1 replica, so the lone pod stayed a standby and never served
(`psql: password authentication failed`). Bump single-replica CI pools to 2.

### DIFF
--- a/pkg/testutil/envtest.go
+++ b/pkg/testutil/envtest.go
@@ -225,8 +225,10 @@ func startManager(t testing.TB, ctx context.Context, mgr manager.Manager) <-chan
 		}
 	}()
 
-	// Wait for cache to sync
-	if !mgr.GetCache().WaitForCacheSync(ctx) {
+	// Wait for cache to sync. WaitForCacheSync returns true immediately when
+	// there are no informers to wait on (e.g. an empty scheme), so also check
+	// the context: a cancelled/expired context means startup did not complete.
+	if !mgr.GetCache().WaitForCacheSync(ctx) || ctx.Err() != nil {
 		t.Fatal("Cache failed to sync")
 	}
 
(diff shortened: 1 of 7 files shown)


Verolop (approved):
LGTM! Thank you 🙏 
Before merging, can you please trigger /e2e on this PR? The visible PR checks are green, but the pull request workflow does not include the e2e matrix

PR #527, opened 2026-06-01
fix(release): resolve goreleaser dist conflict

niconosenzo:
GoReleaser failed with "dist is not empty, remove it before running goreleaser or use the --clean flag" because the release job downloaded install manifests into dist/ before invoking goreleaser.

Download manifests into manifests/ instead, point the release extra_files glob at manifests/install*.yaml, and add --clean so goreleaser starts from a clean dist/. The manifests live outside dist/ so --clean does not delete them.


Verolop (approved):
lgtm! thank you

PR #528, opened 2026-06-01
fix(shard): make example clusters startable out of the box

niconosenzo:
Currently some example manifests used for dev do not scale up, this was surfaced while [repairing the E2E suite](https://github.com/multigres/multigres-operator/pull/524) (where they were worked around in the test harness rather than fixed).

## Changes: 
1. Numeric `RunAsUser` for pool containers: `buildContainerSecurityContext` set `runAsNonRoot=true` but only set a numeric `RunAsUser`/`RunAsGroup` when the pool specified fsGroup. With no fsGroup (the default), the pgctld/multigres images declare `USER postgres` (a name, not a uid), so kubelet cannot verify the container is non-root and refuses to start it with `CreateContainerConfigError`. Pods never leave Pending.
Always set a numeric uid, defaulting to DefaultPostgresUID (999, the postgres user in those images) and still honoring fsGroup as an override. Restores the pre-refactor behavior for the default path.

2. Single-cell pools default to 2 replicas: The default durability policy `AT_LEAST_2` needs 2 poolers *in total* (across all cells) before multiorch can elect a primary. A pool that spans one cell with the per-cell default of 1 yields a single pooler, so it stays an unpromotable standby and never serves (it will hit `psql: password authentication failed`).
When replicasPerCell is omitted, default it to 2 only for a single-cell pool, multi-cell pools already reach 2 total at 1 per cell. The decision uses the effective cell count (the pool's own cells, or all cluster cells when not yet materialized) so it is correct on the dynamic webhook-defaulter path too. DefaultPoolReplicas PerCell stays 1.

Dev clusters workflow fixes:
1. Modify the Makefile kind workflow to pipe `docker save` into `ctr import` without --all-platforms to import just the host platform that is present, and avoid `kind load docker-image`, which runs `ctr import --all-platforms --digests`. With Docker's containerd image store the local image is a multi-platform OCI index, but `docker save` only emits the host platform's blobs, so the import fails with "content digest <sha> not found".

2. Add a `kind-label-nodes` target to the Makefile to label the kind nodes with the single-zone topology labels samples select on.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -53,7 +53,9 @@ endif
 # Be aware that the target commands are only tested with Docker which is
 # scaffolded by default. However, you might want to replace it to use other
 # tools. (i.e. podman)
+# Defaults to docker when unset
 CONTAINER_TOOL ?= docker
+CONTAINER_TOOL := $(if $(strip $(CONTAINER_TOOL)),$(CONTAINER_TOOL),docker)
 
 # Kind cluster name for local development
 KIND_CLUSTER ?= multigres-operator-dev
@@ -369,6 +371,14 @@ define kind-install-crds
 		KUBECONFIG=$(KIND_KUBECONFIG) $(KUBECTL) apply --server-side -f -
 endef
 
+# Load a single image into every node of the kind cluster.
+define kind-load-image
+	@echo "  Loading $(1)..."
+	@for node in $$($(KIND) get nodes --name $(KIND_CLUSTER)); do \
+		$(CONTAINER_TOOL) save $(1) | $(CONTAINER_TOOL) exec -i $$node ctr -n k8s.io images import -; \
+	done
+endef
+
 ##@ Kind Cluster (Local Development)
 
 .PHONY: kind-up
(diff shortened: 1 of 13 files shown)


Makefile
```diff
@@ -369,6 +369,14 @@ define kind-install-crds
 		KUBECONFIG=$(KIND_KUBECONFIG) $(KUBECTL) apply --server-side -f -
 endef
 
+# Load a single image into every node of the kind cluster.
+define kind-load-image
+	@echo "  Loading $(1)..."
+	@for node in $$($(KIND) get nodes --name $(KIND_CLUSTER)); do \
+		docker save $(1) | docker exec -i $$node ctr -n k8s.io images import -; \
```
Verolop: this uses `docker save` directly, but the Makefile still lets callers set `CONTAINER_TOOL` for builds and pulls. If someone uses `CONTAINER_TOOL=podman`, the image may exist only in Podman’s store, so Docker won’t be able to save it here. Wondering if this should use `$(CONTAINER_TOOL) save`, or should we call out that these kind targets require Docker?

Verolop (commented):
Nice! Other than my comment, since your pr description mentions that the e2e suite was working around the fsGroup / numeric UID issue and the single-replica default, could you remove or adjust those workarounds here too? Otherwise, I thiiiiink e2e can still pass even if the new defaults regress later (might be wrong, though). While you're there, after these changes land, the samples README will still say pools default to `replicasPerCell: 1`. Can be addressed in a follow up, if you prefer.

Makefile
niconosenzo: Good catch!, we do it actually in the [contributing guide](https://github.com/multigres/multigres-operator/blob/main/CONTRIBUTING.md), we can likely make it more evident by adding a macro to check prereqs in the makefile during kind-* target calls.

niconosenzo: Added a `CONTAINER_TOOL` variable that will default to docker, I thought on not adding that on the documentation for now, at least until all the targets that will call it are tested with other tools (i.e. podman).

Verolop (approved):
lgtm!

PR #531, opened 2026-06-08
fix(e2e): support multi-doc YAML in e2e framework loader

niconosenzo:
`minimal.yaml `prepends a Secret for `kubectl apply-f` UX, but `decoder.Decode` only reads the first doc and returned `*v1.Secret,` not `*MultigresCluster`, causing a panic.

`LoadYAML` now iterates all docs via `YAMLOrJSONDecoder` and returns `[]runtime.Object`. MustLoad* functions scan for the target type.

### DIFF
--- a/test/e2e/framework/loader.go
+++ b/test/e2e/framework/loader.go
@@ -3,13 +3,16 @@
 package framework
 
 import (
+	"bytes"
 	"fmt"
+	"io"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


niconosenzo:
[Test run](https://github.com/niconosenzo/multigres-operator/actions/runs/27134499509)

Verolop (approved):

PR #533, opened 2026-06-09
feat(poolers): mark dead poolers with SHUTDOWN lifecycle

niconosenzo:
When a pooler pod dies ungracefully (SIGKILL, crash, abrupt scale-down) it never announces its own shutdown, so its topology entry lingers with a live lifecycle and no backing pod. The orchestrator keeps treating it as a cohort member.

Detect these entries in the shard reconcile loop and mark them LIFECYCLE_SHUTDOWN via UpdateMultiPoolerFields. The orchestrator's pooler watcher keys on this transition to tear down the health stream and clear the pooler from the cohort. This replaces the previous behavior of hard-deleting orphaned entries (PrunePoolers -> MarkDeadPoolers), which left the orchestrator seeing them as unreachable health-stream failures.

Also, introduce a safeguard (`isLifecycleShutdown(p.MultiPooler)`) to avoid including a pooler being shutdown in `GetPoolerStatus` or `FindPrimaryPooler` functions.


pkg/data-handler/topo/pooler.go
```diff
@@ -192,28 +207,54 @@ func PrunePoolers(
...
+
+			hostname := p.GetHostname()
+			if hostname == "" && p.Id != nil {
+				hostname = p.Id.Name
+			}
+
+			if _, err := store.UpdateMultiPoolerFields(ctx, p.Id,
+				func(mp *clustermetadatapb.MultiPooler) error {
+					mp.LifecycleStatus = &clustermetadatapb.PoolerLifecycle{
+						Status:  clustermetadatapb.PoolerLifecycleStatus_LIFECYCLE_SHUTDOWN,
```
dweitzman, line 238: Could add a TODO here to consider making a new lifecycle status like, `DEPROVISIONED` or something to reflect that the shutdown wasn't graceful. It wouldn't impact orch's behavior, but could be useful to something for logging or for people browsing etcd and wondering what happened to a pooler

pkg/data-handler/topo/pooler_test.go
```diff
@@ -106,6 +106,41 @@ func TestFindPrimaryPooler(t *testing.T) {
...
+
+	t.Run("skips a shut-down primary", func(t *testing.T) {
+		t.Parallel()
+		_, factory := memorytopo.NewServerAndFactory(context.Background(), "cell1")
+		store := topoclient.NewWithFactory(
+			factory, "", []string{""}, topoclient.NewDefaultTopoConfig(),
+		)
+		defer func() { _ = store.Close() }()
+
+		ctx := context.Background()
```
dweitzman: Nit: could use `t.Context()`

dweitzman (approved):
Nice!

pkg/data-handler/topo/pooler_test.go
niconosenzo: It is a pattern widely used on the tests for this repository, will open a followup PR to change them all.

PR #537, opened 2026-06-29
refactor(tablegroup): make shard reconciliation observed-state truthful

Verolop:
## Description

this PR refactors the TableGroup reconciler around the lifecycle it owns: applying desired child Shards, retiring removed or replaced children through the existing drain handshake, and publishing aggregate status from observed child state.

The previous implementation kept child apply, orphan cleanup, child re-listing, status aggregation, event emission, and requeue decisions in one long reconcile flow. That made the controller’s state transitions implicit, especially around removed children and same-reconcile child updates. In those cases, TableGroup could leave stale parent status visible or count child state that no longer represented the current desired generation.

The new structure keeps TableGroup’s responsibility unchanged. It still applies owned Shard specs with Server-Side Apply and it still preserves the PendingDeletion / ReadyForDeletion cleanup flow. The refactor makes the lifecycle boundaries explicit in code and tightens status aggregation so the parent reports convergence honestly while children are draining, terminating or catching up.

## Main Changes

- Split the TableGroup reconcile flow into package-local phases for parent lifecycle handling, child apply/cleanup, status aggregation, status patching, and cleanup requeue decisions.

- Replaced the previous normal-path child re-list with one observed child snapshot, so status is computed from child state the controller actually saw.

- Preserved the existing `PendingDeletion` / `ReadyForDeletion` cleanup handshake while preventing removed, replaced, or terminating children from being applied over or counted active.

- Counted readiness only from desired child Shards whose observed status has caught up to the generation applied by this reconcile.

- Added focused unit and envtest coverage for child cleanup, same-name replacement, terminating-child handling, generation-gated readiness, and stable convergence.

## Testing

- Added coverage around the lifecycle and status invariants that make this refactor risky.

- Status aggregation is tested directly: total comes from spec, readiness and degradation come from observed desired children, cleanup children do not make the parent look ready, pending cleanup forces `Progressing`, and readiness is gated on observed generation.

- Child cleanup is tested at the lifecycle boundaries: removed Shards are annotated once, existing annotations are preserved, children are not deleted before `ReadyForDeletion`, same-name replacements do not apply over children already in cleanup, and terminating children are waited on rather than restamped.

- Full reconcile tests verify the controller lists child Shards once on the normal path, publishes cleanup status before requeueing, requeues while parent cleanup is waiting on a terminating child, and does not flap an already-Healthy TableGroup.

- Envtest coverage exercises convergence against a real apiserver/cache, including initial child creation, stable terminal status, add/remove reconciliation, retained child cleanup, same-name replacement after deletion, and reconvergence after the replacement child reports Healthy.

### DIFF
--- a/pkg/cluster-handler/controller/tablegroup/reconcile_shards.go
+++ b/pkg/cluster-handler/controller/tablegroup/reconcile_shards.go
@@ -1,0 +1,148 @@
+package tablegroup
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+
+	multigresv1alpha1 "github.com/multigres/multigres-operator/api/v1alpha1"
+)
+
+// stepListChildShards reads the children once for the normal reconcile path.
+// Later steps apply desired children and prune orphans against this same
+// snapshot, preserving each child's observed .Status for status aggregation.
+func stepListChildShards(ctx context.Context, rc *reconcileContext) (stepResult, error) {
+	observed := &multigresv1alpha1.ShardList{}
+	if err := rc.r.List(ctx, observed, childShardSelector(rc.tg)...); err != nil {
+		return stepResult{}, newStepError(
+			fmt.Errorf("failed to list shards for pruning: %w", err),
+			"Warning",
+			"CleanUpError",
+			fmt.Sprintf("Failed to list shards for pruning: %v", err),
+		)
+	}
+
+	rc.observedShards = observed
+	return continueStep(), nil
+}
+
+// stepApplyDesiredShards applies each desired Shard and records the names and
+// generations produced by the apiserver. Later status aggregation uses those
+// generations to avoid treating a just-updated child as ready before the child
+// controller observes its new spec.
+func stepApplyDesiredShards(ctx context.Context, rc *reconcileContext) (stepResult, error) {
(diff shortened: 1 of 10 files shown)


pkg/cluster-handler/controller/tablegroup/reconcile_shards.go
```diff
@@ -0,0 +1,148 @@
...
+
+	rc.observedShards = observed
+	return continueStep(), nil
+}
+
+// stepApplyDesiredShards applies each desired Shard and records the names and
+// generations produced by the apiserver. Later status aggregation uses those
+// generations to avoid treating a just-updated child as ready before the child
+// controller observes its new spec.
+func stepApplyDesiredShards(ctx context.Context, rc *reconcileContext) (stepResult, error) {
```
niconosenzo, line 38: Is there any chance we end up trying to re-add a shard that has been previously annotated with `PendingDeletion` ?

niconosenzo:
Awesome work! Left a small comment, will stamp it after knowing your thoughts.

pkg/cluster-handler/controller/tablegroup/tablegroup_controller.go
```diff
@@ -54,299 +54,70 @@ func (r *TableGroupReconciler) Reconcile(
...
 
-	// If this TableGroup is pending deletion (from MultigresCluster pruning),
-	// propagate PendingDeletion to all child Shards and wait.
-	if tg.Annotations[multigresv1alpha1.AnnotationPendingDeletion] != "" {
-		return r.handlePendingDeletion(ctx, tg)
+	// The order preserves the controller lifecycle: fetch the parent, handle
+	// parent deletion before normal work, read the child snapshot once, apply the
+	// desired children, retire orphans through their drain handshake, then publish
+	// parent status from the observed child state.
+	rawSteps := []step{
```
niconosenzo, line 67: This is great. Should we make this list of steps a contract that every controller will eventually implement ?

niconosenzo (approved):

PR #539, opened 2026-07-07
chore: Reduce certificate lifetime to 14 days, clean up subject/SANs

jnschaeffer:
This PR reduces the duration of an issued certificate from 5 years to 14 days, fixes a typo in the subject name, and removes logic that modifies the DNS SANs based on whether the DNS name starts with "db.".

### DIFF
--- a/pkg/cluster-handler/controller/multigrescluster/certificate.go
+++ b/pkg/cluster-handler/controller/multigrescluster/certificate.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"strings"
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	apimeta "k8s.io/apimachinery/pkg/api/meta"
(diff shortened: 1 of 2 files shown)


pkg/cluster-handler/controller/multigrescluster/certificate.go
```diff
@@ -47,11 +46,7 @@ func buildCertificate(
 ) (*unstructured.Unstructured, error) {
 	cn := cluster.Spec.CertCommonName
 
-	// Build the secondary SAN by stripping the "db." prefix if present.
 	dnsNames := []any{cn}
```
Verolop, line 49: Can you please update `certificate_test.go` to match this? `TestBuildCertificate` still expects the stripped secondary SAN (`abc123.supabase.red`) here, plus the old `Delware` subject string. 

Also, can you confirm that dropping the stripped SAN is safe? The subject spelling fix seems fine, but removing this SAN could break any client still connecting via the non-`db.` hostname.

pkg/cluster-handler/controller/multigrescluster/certificate.go
```diff
@@ -22,12 +21,12 @@ const (
 	// CertIssuerName is the cert-manager ClusterIssuer used for TLS certificates.
 	CertIssuerName = "supabase-issuer"
 
-	// CertDuration is the certificate duration (5 years), matching non-HA projects.
-	CertDuration = "44640h0m0s"
+	// CertDuration is the certificate duration.
+	CertDuration = "336h0m0s"
```
niconosenzo, line 25: Wondering if we should separate the constructors or use different duration for, e.g., external facing certificates (multigateway) and component-to-component certificates.

PR #540, opened 2026-07-08
feat: wire internal gRPC TLS

niconosenzo:
## Description

Wire cert-manager-backed TLS/mTLS for managed Multigres workload gRPC
listeners and multipooler-facing clients, using the TLS support added in
multigres/multigres#826.

The operator issues separate certificates and Secrets for multiadmin,
multigateway, multiorch, and multipooler. Their primary logical identities use:

`<component>.<cluster>.<namespace>.multigres.internal`

Public PostgreSQL TLS remains configured separately through the existing
`generated-certs` Secret and `certCommonName`.

### Internal workload traffic policy

When TLS is enabled:

- **multipooler**
  - Its gRPC listener uses its multipooler certificate.
  - The listener requires and verifies client certificates through `--grpc-ca`.

- **multiorch**
  - Its gRPC listener requires and verifies client certificates.
  - multiorch presents its component certificate when calling multipooler.
  - multipooler server certificates are verified against
    `multipooler.<cluster>.<namespace>.multigres.internal`.
  - `--multipooler-grpc-require-tls` prevents plaintext fallback.

- **multiadmin**
  - Its gRPC listener requires and verifies client certificates.
  - multiadmin presents its component certificate when calling multipooler.
  - multipooler server certificates are verified against the logical
    multipooler identity.
  - `--multipooler-grpc-require-tls` prevents plaintext fallback.

- **multigateway**
  - Its gRPC listener is internal-only and requires and verifies client
    certificates through `--grpc-ca`.
  - multigateway presents its component certificate when calling multipooler.
  - multipooler server certificates are verified against the logical
    multipooler identity.
  - `--multipooler-grpc-require-tls` prevents plaintext fallback.

The projected certificate secrets use file modes compatible with non-root
workloads. Stale cert-manager Certificates and their generated Secrets are
removed when certificate names change, with cleanup remaining retryable.

### Current compatibility constraints

Upstream Multigres currently reuses the `--multipooler-grpc-*` transport
configuration for gateway-to-gateway cancel forwarding. multiadmin also reuses
its multipooler transport configuration for RPCs to other component types.

Because those clients currently use the multipooler server-name override, the
multigateway certificate temporarily includes the logical multipooler SAN.
Destination-specific upstream TLS settings are needed before that compatibility
SAN can be removed.

### Enablement

Internal workload TLS is currently enabled when
`MultigresCluster.spec.certCommonName` is set. Clusters without
`certCommonName` remain unchanged.

Internal and public certificates currently use the same configured
`supabase-issuer`. Separating internal TLS enablement and issuer configuration
from the public PostgreSQL certificate remains follow-up work.

### DIFF
--- a/pkg/cluster-handler/controller/multigrescluster/certificate.go
+++ b/pkg/cluster-handler/controller/multigrescluster/certificate.go
@@ -6,8 +6,12 @@ import (
 	"fmt"
 	"strings"
 
+	apiequality "k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	apimeta "k8s.io/apimachinery/pkg/api/meta"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
@@ -36,6 +40,14 @@ var certGVK = schema.GroupVersionKind{
 	Kind:    "Certificate",
 }
 
+type certSpec struct {
+	name       string
+	secretName string
+	commonName string
+	dnsNames   []any
+	usages     []any
+}
+
 // buildCertificate constructs an unstructured cert-manager Certificate for the
 // multigateway TLS certificate. The Certificate spec matches what non-HA
 // projects use, with the supabase-issuer ClusterIssuer.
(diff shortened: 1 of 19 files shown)


pkg/resource-handler/controller/cell/multigateway.go
```diff
@@ -237,17 +245,46 @@ func BuildMultiGatewayDeployment(
...
 		podSpec.Containers[0].Args = append(podSpec.Containers[0].Args,
 			"--pg-tls-cert-file", tlsCertFile,
 			"--pg-tls-key-file", tlsKeyFile,
+			"--grpc-cert", internalTLSCertFile,
+			"--grpc-key", internalTLSKeyFile,
+			"--grpc-server-ca", internalTLSCAFile,
+			"--multipooler-grpc-cert", internalTLSCertFile,
+			"--multipooler-grpc-key", internalTLSKeyFile,
+			"--multipooler-grpc-ca", internalTLSCAFile,
+			"--multipooler-grpc-server-name",
```
Verolop, line 284: Could you split the gateway-to-gateway TLS settings from the multipooler settings and use `multigateway.<CN>` as the server name? Cancel forwarding currently reuses this flag, so the gateway certificate won’t verify against `multipooler.<CN>`. Any other alternative works too.

pkg/cluster-handler/controller/multigrescluster/certificate.go
```diff
@@ -147,20 +225,50 @@ func (r *MultigresClusterReconciler) deleteOwnedCertificates(
...
 			continue
 		}
-		if cert.GetName() == keepName {
-			continue
+		if keepNames != nil {
+			if _, ok := keepNames[cert.GetName()]; ok {
+				continue
+			}
 		}
 		if err := r.Delete(ctx, cert); err != nil &&
```
Verolop, line 337: cert-manager keeps Secrets by default, so CN changes currently leave old private keys behind, can you delete the generated Secret along with the stale Cert?

pkg/cluster-handler/controller/multigrescluster/certificate.go
```diff
@@ -115,30 +168,55 @@ func (r *MultigresClusterReconciler) reconcileCertificate(
...
+	// Clean up stale Certificates owned by this cluster. When the CN changes the
+	// old Certificates would otherwise linger and conflict on Secret names.
+	if err := r.deleteOwnedCertificates(ctx, cluster, certList, keepNames); err != nil {
+		return err
+	}
+
+	for _, cert := range desiredCerts {
+		// Skip the Patch entirely when the live spec already matches the
+		// desired spec, so a no-op reconcile doesn't churn resourceVersion.
+		if existing := findCertificateByName(certList, cert.GetName()); existing != nil &&
```
Verolop, line 231: I think it's worth including `isOwnedBy(existing, cluster)` in this check too. Otherwise, a matching spec skips the apply when the owner reference is missing, and the Certificate stays unmanaged.

Verolop (commented):
left a couple of comments for minor naming & cleanup.
The main part looks great to me 🚀

Verolop (approved):
lgtm!

PR #542, opened 2026-07-15
fix(shard): decouple fsGroup from pool runtime identity

Verolop:
## Description:

this PR fixes pool pod security context handling so `fsGroup` is treated as pod-level filesystem group access, not as the container process UID/GID. It preserves image-provided runtime identity by default and adds explicit UID/GID configuration for cases that need it.

## Main Changes:

- Adds independent `runAsUser` and `runAsGroup` fields to container config.
- Stops deriving container `runAsUser` / `runAsGroup` from `fsGroup`.
- Keeps `fsGroup` scoped to `PodSecurityContext.fsGroup`.
- Merges runtime identity fields through shard template/override resolution.
- Validates final shared-PGDATA UID compatibility after resolution.
- Regenerates deepcopy and CRD schemas for the new API fields.

## Testing:

Tests cover the behaviour at each layer where the bug could regress: API/schema generation, template and override resolution, admission-time resolved validation, and final pod rendering. Unit tests verify that `fsGroup` no longer changes container UID/GID, explicit identities are preserved, custom images can retain image identity, and runtime identity changes affect pod hashes. Resolver tests cover inheritance and partial override behaviour, including valid and invalid final shared-PGDATA identities. CEL/envtest coverage verifies that CRD schema validation accepts partial identity shapes while leaving final compatibility checks to resolved validation.


niconosenzo (approved):
lgtm!

PR #544, opened 2026-07-22
chore(multigres): bump images and module to 54b4c18

Verolop:
## Description

this PR bumps the Multigres images and Go module. To remain compatible with upstream, primary detection now uses `routing_state.role` instead of the deprecated `Multipooler.Type`, and `LIFECYCLE_QUARANTINED` is translated at the topology boundary to the operator’s existing `DRAINED` value only to preserve compatibility with the current operator model. The drain logic itself is not changed or redesigned here yet.

## Main changes
- Bump Multigres and pgctld images to sha-54b4c18, multiadmin-web to sha-95eb32, bump Multigres Go module as well.
- Replace deprecated PoolerType reads with routing_state.role.
- Preserve the REPLICA fallback for missing routing state during mixed-version rollouts.
- Map LIFECYCLE_QUARANTINED to the existing DRAINED workflow so unusable poolers trigger replacement capacity.
- Update affected test fixtures to use routing state.

## Testing

Updated unit coverage for routing-state-based primary discovery, quarantined and shut-down pooler handling, and the replica fallback used during mixed-version rollouts.


mkindahl (approved):
LGTM. I checked that all uses of PoolerType was removed and uses RoutingRole instead.

PR #545, opened 2026-07-22
feat(api/shard): add backup encryption config

niconosenzo:
BackupConfig gets an optional Encryption block (secretName). The Database topology registration sets require_initial_repo_encryption so poolers refuse to start without the cipher key mounted.

When backup encryption is enabled, reconcile a cipher key Secret, mount it read-only into the multipooler container, and pass --pgbackrest-cipher-key-file.


pkg/data-handler/topo/topology.go
```diff
@@ -47,6 +47,8 @@ func RegisterDatabaseFromSpec(
 		BootstrapDurabilityPolicy: bootstrapDurabilityPolicy,
 	}
 
+	requireInitialRepoEncryption := backup != nil && backup.Encryption != nil
```
Verolop, line 50: Not blocking for the initial db-level flow, but BackupConfig can also be overridden at the table group or shard level. In those cases, the cipher key reaches the pooler while r`equire_initial_repo_encryption` remains false in topo.

Verolop (approved):

pkg/data-handler/topo/topology.go
niconosenzo: Good catch, wondering in which cases we would enable encryption downstream and not in the parent CR, asking mainly to understand if we should explicitly account for that.

niconosenzo: We can add a safeguard for this in the feature if needed, thanks for the comment!.

PR #548, opened 2026-07-28
feat(tls): introduce dedicated internal tls setup flag

niconosenzo:
Add internalTLS flag, so we can trigger x-component TLS setup independently from public TLS.

`InternalTLS` now controls:

- Internal cert-manager Certificates for multiadmin, multigateway, multiorch, and multipooler
- Internal Secret volumes and mounts
- gRPC server TLS/mTLS flags
- Multipooler client TLS flags and server-name verification
- Removal of stale internal Certificates and generated Secrets when internal mTLS is disabled

`certCommonName` now controls only:

- The public multigateway PostgreSQL Certificate
- The generated-certs Secret mount
- --pg-tls-cert-file and --pg-tls-key-file


Verolop (approved):
lgtm! 🔒

PR #549, opened 2026-07-28
fix(ci): consume attested full-sha nightlies

Verolop:
## Description

this pr updates the nightly compatibility canary to consume the immutable, attested images produced by the Multigres nightly build.

## Main changes

- Reads the source SHA from the latest successful upstream nightly run.
- Uses full-SHA image tags for verification, e2e execution, and failure reporting.
- Verifies provenance directly from each image’s OCI registry bundle.
- Keeps permissions minimal while preserving ancestry validation, digest pinning, and tracking issue behavior.

## Testing

Validated the workflow with `actionlint`, `zizmor`, and `git diff --check`. The updated resolution and verification flow was also checked against a successful upstream nightly build with published full-SHA images and provenance attestations.

### DIFF
--- a/.github/workflows/nightly-compatibility.yaml
+++ b/.github/workflows/nightly-compatibility.yaml
@@ -88,32 +88,15 @@ jobs:
         run: |
           sha="$INPUT_SHA"
           if [ -z "$sha" ]; then
-            run_id="$(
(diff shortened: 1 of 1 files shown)


rafael (approved):
Stamping

PR #551, opened 2026-07-28
fix(e2e): handle digest-pinned canary images

Verolop:
## Description

this pr fixes two e2e harness failures exposed by the digest-pinned nightly compatibility canary.

## Main changes

- Lets Kubernetes pull digest-pinned runtime images directly instead of passing digest-only references through Kind’s Docker archive import.
- Keeps locally tagged images on the existing Kind load path.
- Makes the webhook removal fixture valid by supplying the required cell topology.
- Adds coverage for excluding digest references from Kind loading.

## Testing

Ran the e2e framework unit tests with the `e2e` build tag and compiled the webhook e2e package. Also verified formatting and whitespace with `gofmt` and `git diff --check`.

### DIFF
--- a/test/e2e/framework/cluster.go
+++ b/test/e2e/framework/cluster.go
@@ -80,16 +80,32 @@ func (c *Cluster) CRClient() (client.Client, error) {
 	return client.New(c.RestCfg, client.Options{Scheme: scheme})
 }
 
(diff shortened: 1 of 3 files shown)


rafael (approved):
LGTM

PR #554, opened 2026-07-30
feat(postgres-config): operator-owned postgresql.conf per shard

GuptaManan100:
## What

Make the operator own each shard's PostgreSQL configuration (`postgresql.conf`), rendered as a layered config into an operator-owned ConfigMap and rolled out through the existing primary-last restart. Precedence, lowest → highest:

1. **Operator baseline** — a static `postgresql.conf` template the operator owns end-to-end (SSL, locales, logging, `wal_level = logical`, and conservative small-instance defaults).
2. **Resource-derived sizing** — pgtune-style GUCs computed from the shard's pool resources (`shared_buffers = mem/4`, `effective_cache_size`, `work_mem`, `max_parallel_workers*`, …) plus disk-based WAL sizing (`max_wal_size`, `wal_keep_size`, …).
3. **Legacy `postgresConfigRef`** — the existing ConfigMap-ref content, now **deprecated** in favor of the inline map.
4. **Inline `spec.postgresConfig`** — a user-supplied `map[string]string` of GUC overrides, the intended long-term surface.

The result is written to an operator-owned `<shard>-postgres-config` ConfigMap and mounted into the `postgres` sidecar via `POSTGRES_INITDB_EXTRA_CONF`. A content hash of the rendered config drives the existing drain → primary-last restart state machine, so config changes roll out the same way resource changes already do.

Config is **shard-level, not per-pool**: every pool in a shard gets identical config, which keeps primaries and replicas uniform across failover. Where sizing needs a single basis, per-pool resources are reduced to the shard — **max(memory, CPU)** across pools (so replication-sensitive settings stay valid on every pod) and **min(disk)** (so WAL budgeting never overfills the smallest volume). `cluster_name` is rendered from the cluster/database/tablegroup/shard identity so `ps`/log output distinguishes shards.

## Validation (admission)

`spec.postgresConfig` GUCs are validated at admission against a 382-entry catalog generated from PostgreSQL's `guc_tables.c` (REL_17_5): unknown parameter names and gross type mismatches are rejected. Additionally, **operator-managed GUCs are rejected** rather than silently ignored:

- pgctld pins connection/path params on the command line (`port`, `listen_addresses`, `unix_socket_directories`, `data_directory`, `hba_file`, `ident_file`), so a config-file value would be a no-op;
- `wal_level` must stay `logical` for multigres replication.

## Status

`status.postgresConfig` reports rollout state so callers can poll without inspecting pods: `{ inProgress, lastAppliedAt, error }`.

- `inProgress` is **content-based** — it compares the config effective on the pods against the desired render, so it correctly reflects a `postgresConfigRef` ConfigMap edit or a new operator baseline on upgrade, neither of which bumps the shard `generation`. (Freshness for a spec change is the shard's top-level `status.observedGeneration`.)
- A rollout is reported **settled only when there is neither config drift nor an apply failure** — a pod already carrying the desired hash that is crash-looping (the signature of a GUC value Postgres rejects at startup) keeps the rollout unsettled instead of falsely reporting "applied". The gate is keyed on the desired hash, so an unrelated crash-loop on a stale-hash pod doesn't false-flag.

## Testing

- Unit + envtest integration green, incl. a webhook integration test that rejects an unknown GUC through the real mutating + validating webhooks.
- `TestPostgresConfigManagement` (e2e, kind) brings up a sized cluster with an inline map and a legacy ref, then asserts **effective** values via `SHOW <param>` through the gateway: baseline `wal_level=logical`, `shared_buffers=128MB` (from a 512Mi limit), inline `max_connections=150`, map-over-ref `work_mem=16MB`, ref-honored `random_page_cost=2.5`, and the per-shard ConfigMap.

## Dependency / known limitation

Applying config **changes** to an already-running cluster depends on an upstream pgctld change. Today pgctld reads `POSTGRES_INITDB_EXTRA_CONF` only at **initdb**, so:

- **Initial** config (baseline + sizing + overrides at cluster creation) already takes effect.
- **Changing** config on a running cluster rolls the pods, but the new values are not re-read until pgctld includes the mounted file live on restart (`include_if_exists`). The e2e subtest `changing spec.postgresConfig takes effect on the running cluster` is present and will pass once that lands; the managed-GUC and apply-health guardrails above also become fully active then.

## Follow-up (separate PR)

**Reload vs. restart classification** is intentionally **out of scope here** and will be a separate PR. Today every config change takes the full rolling-restart path. The follow-up will classify changed GUCs by `pg_settings.context` and apply `sighup`-context changes via a **verified** reload (no pod restart), falling back to restart otherwise. It needs the observability half to confirm a reload actually took effect (a verifying pgctld reload RPC, or an operator-side check) — this PR deliberately makes the status signal content-based and observation-driven so that reload path plugs into the same "effective == desired" loop without an API change.


niconosenzo (approved):
🔥 !

PR #555, opened 2026-07-31
chore(toposerver): always delete topo PVCs with cluster

niconosenzo:
Topo etcd data is coordination state rebuilt on cluster creation, so retaining its PVCs after a MultigresCluster is deleted only leaves orphaned volumes behind.

What this change introduce, gate TopoServer deletion with a finalizer so rebuildable etcd volumes are removed when the owning MultigresCluster is terminating or gone. Reconcile deletion-timestamp transitions explicitly and preserve data-plane PVCs under Retain.

### DIFF
--- a/pkg/resource-handler/controller/toposerver/statefulset.go
+++ b/pkg/resource-handler/controller/toposerver/statefulset.go
@@ -18,7 +18,7 @@ import (
 
 const (
 	// ComponentName is the component label value for toposerver resources
-	ComponentName = "toposerver"
+	ComponentName = metadata.ComponentTopoServer
 
 	// DefaultReplicas is the default number of etcd replicas
 	DefaultReplicas int32 = 3
--- a/test/e2e/dedicated/deletion/deletion_test.go
+++ b/test/e2e/dedicated/deletion/deletion_test.go
@@ -4,6 +4,7 @@ package deletion_test
 
 import (
 	"context"
+	"strings"
 	"testing"
 	"time"
 
@@ -30,6 +31,11 @@ func TestClusterDeletion(t *testing.T) {
 	// Load and create the minimal sample.
 	cr := framework.MustLoadCluster("config/samples/minimal.yaml", ns)
 	cr.Name = "delete-me" // distinct name for clarity in logs
+
+	cr.Spec.PVCDeletionPolicy = &multigresv1alpha1.PVCDeletionPolicy{
+		WhenDeleted: multigresv1alpha1.RetainPVCRetentionPolicy,
+		WhenScaled:  multigresv1alpha1.RetainPVCRetentionPolicy,
+	}
 	if err := c.Create(ctx, cr); err != nil {
 		t.Fatalf("create MultigresCluster: %v", err)
 	}
@@ -83,4 +89,39 @@ func TestClusterDeletion(t *testing.T) {
 		func(l *corev1.ServiceList) int { return len(l.Items) },
 		"Services", 3*time.Minute,
 	)
+
+	// Verify topo (etcd) PVCs are cleaned up. The MultigresCluster cleanup
+	// finalizer deletes topo PVCs regardless of the configured PVCDeletionPolicy
+	// (which this test forces to Retain above), so
+	// PVCs from the global topo StatefulSet should not survive the cluster.
+	framework.WaitForEmpty(t, c, ns,
+		&corev1.PersistentVolumeClaimList{},
+		func(l *corev1.PersistentVolumeClaimList) int {
+			count := 0
+			for _, pvc := range l.Items {
+				if strings.HasPrefix(pvc.Name, "data-"+cr.Name+"-") &&
+					strings.Contains(pvc.Name, "-topo-") {
+					count++
+				}
+			}
+			return count
+		},
+		"topo PVCs", 3*time.Minute,
+	)
+
+	// Verify data (non-topo) PVCs are retained under the Retain policy forced
+	// above, only topo PVCs are force-deleted by cluster cleanup.
+	pvcList := &corev1.PersistentVolumeClaimList{}
+	if err := c.List(ctx, pvcList, client.InNamespace(ns)); err != nil {
+		t.Fatalf("list PVCs: %v", err)
+	}
+	nonTopoCount := 0
+	for _, pvc := range pvcList.Items {
+		if !strings.Contains(pvc.Name, "-topo-") {
+			nonTopoCount++
+		}
+	}
+	if nonTopoCount == 0 {
+		t.Fatalf("expected at least one non-topo (data) PVC to survive cluster deletion under Retain, found none")
+	}
 }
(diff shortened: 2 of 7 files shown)


pkg/resource-handler/controller/toposerver/statefulset.go
```diff
@@ -39,7 +39,14 @@ func BuildStatefulSet(
...
 	scheme *runtime.Scheme,
 ) (*appsv1.StatefulSet, error) {
-	pvcPolicy := toposerver.Spec.PVCDeletionPolicy
+	// Topo etcd data is coordination state that gets rebuilt on cluster creation,
+	// so topo PVCs are always removed with the StatefulSet regardless of
+	// PVCDeletionPolicy: WhenDeleted is forced to Delete here and does not honor
+	// an explicit Retain anywhere in the hierarchy. WhenScaled still follows the
+	// configured policy (pvc.BuildRetentionPolicy).
+	retentionPolicy := pvc.BuildRetentionPolicy(toposerver.Spec.PVCDeletionPolicy)
+	retentionPolicy.WhenDeleted = appsv1.DeletePersistentVolumeClaimRetentionPolicyType
```
Verolop: Could we retain PVCs here and clean them up specifically during cluster deletion? Replacing this StatefulSet while the cluster is running would delete its etcd data and recreate it empty, no? My concern is that whenDeleted means StatefulSet deletion, not MultigresCluster deletion

pkg/cluster-handler/controller/multigrescluster/builders_cell.go
```diff
@@ -21,6 +21,15 @@ func BuildCell(
 	allCellNames []multigresv1alpha1.CellName,
 	scheme *runtime.Scheme,
 ) (*multigresv1alpha1.Cell, error) {
+	localTopo := localTopoSpec.DeepCopy()
+	// Merge hierarchy: Etcd -> local toposerver -> multigrescluster.
+	if localTopo != nil && localTopo.Etcd != nil {
+		localTopo.Etcd.PVCDeletionPolicy = multigresv1alpha1.MergePVCDeletionPolicy(
```
Verolop: Switching the global topo to external currently deletes every TopoServer with the cluster label, including local ones, so with this change, that would also delete their PVCs. Could we scope that cleanup to the global TopoServer first? Could be a separate PR.

test/e2e/dedicated/deletion/deletion_test.go
```diff
@@ -83,4 +84,24 @@ func TestClusterDeletion(t *testing.T) {
 		func(l *corev1.ServiceList) int { return len(l.Items) },
 		"Services", 3*time.Minute,
 	)
+
+	// Verify topo (etcd) PVCs are cleaned up. Topo PVCs default to Delete on
+	// cluster deletion (unlike data volumes, which default to Retain), so
+	// PVCs from the global topo StatefulSet ("data-<cluster>-global-topo-*")
+	// and any local (cell) topo StatefulSets ("data-<cluster>-<cell>-topo-*")
+	// should not survive the cluster.
+	framework.WaitForEmpty(t, c, ns,
```
Verolop, line 97: this test needs a small adjustment to prove the PR is doing the work. The sample already uses `whenDeleted: Delete`, so this passes before the change.

Verolop (commented):
Thanks for tackling the orphaned PVCs. So happy we're finally at this stage 🥳 
I left one lifecycle blocking comment, and a couple of small ones.

pkg/resource-handler/controller/toposerver/statefulset.go
niconosenzo: 🫤 that's correct.

pkg/cluster-handler/controller/multigrescluster/builders_cell.go
niconosenzo: Decided to not rely on the PVCDeletionPolicy for the topo PVCs, if the multigrescluster owner object has been deleted or will be deleted, we get rid of the PVC. Unless we have a reason to keep the volumes after that event ?

pkg/resource-handler/controller/toposerver/reconcile_deletion.go
```diff
@@ -0,0 +1,165 @@
...
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+
+	multigresv1alpha1 "github.com/multigres/multigres-operator/api/v1alpha1"
+	"github.com/multigres/multigres-operator/pkg/util/metadata"
+)
+
+// handleDeletion runs when a TopoServer is being deleted. It is gated by
```
Verolop: When a cluster switches from managed topo to external, this TopoServer is deleted while the cluster is still running. This handler keeps the PVC and removes the finalizer, which is right at that point, but if the cluster is deleted later, there is no TopoServer left to come back and remove that PVC. Kubernetes GC cannot help because the PVC has no cluster owner, and multigres-gc will not see it because it has no orphan-since label.
One way to deal with this is to make the PVC follow the MultigresCluster, not the TopoServer. Add the cluster as a non-controller owner of the PVC, so Kubernetes keeps it while that exact cluster exists and deletes it when the cluster is gone, but if you prefer explicit cleanup, put the finalizer on the MultigresCluster for the same reason.
One e2e test for this path should help too: start with managed topo and Retain, switch to external and confirm the PVC stays, then delete the cluster and confirm the PVC is removed, but this can be a separate pr later.

pkg/resource-handler/controller/toposerver/reconcile_deletion.go
```diff
@@ -0,0 +1,165 @@
...
+			"Retaining topo PVCs; owning MultigresCluster is still alive",
+			"cluster", clusterName,
+		)
+	}
+
+	// Cleanup has been attempted (either path), remove the finalizer so
+	// k8s can finish deleting the TopoServer. Any error above returns
+	// early, leaving the finalizer in place so reconciliation retries.
+	if slices.Contains(toposerver.Finalizers, topoServerFinalizer) {
+		patch := client.MergeFrom(toposerver.DeepCopy())
```
Verolop: small safety note: this patch replaces the whole finalizer list using the object we read earlier. If another finalizer is added at the same time, this patch can accidentally remove it.
You can fetch the latest TopoServer and use Update, or use MergeFromWithOptimisticLock, and make the same change where the finalizer is added.

pkg/resource-handler/controller/toposerver/statefulset.go
Verolop: There’s still one path where etcd data can disappear while the cluster is running. If the StatefulSet uses `whenDeleted: Delete`, this handler keeps the PVC, but Kube deletes it anyway when the StatefulSet goes away. The TopoServer can then come back with a new, empty PVC.
Keep `whenDeleted: Retain` on topo StatefulSets and remove their PVCs only when the MultigresCluster itself is deleted.

pkg/cluster-handler/controller/multigrescluster/builders_cell.go
Verolop: Agreed on deleting the topo PVCs when the MultigresCluster is deleted. I saw that the selector issue I mentioned is separate and already exists today, plus it's only for one specific case, so it does not need to block this PR :D

Verolop (changes requested):
Thanks for reworking this 💚 

I left two lifecycle comments that need addressing before merge. One covers the case where the TopoServer is already gone before the cluster is deleted. The other covers the StatefulSet still deleting the PVC while the cluster is running.

I also left one smaller finalizer safety comment. The global/local selector issue already existed before this change, so I don’t consider that a blocker for this PR.

pkg/resource-handler/controller/toposerver/reconcile_deletion.go
niconosenzo: Thanks for good comments @Verolop !

Verolop (approved):
🚀

PR #556, opened 2026-08-01
feat(images): resolve component imgs at reconcile time; paced fleet rollout

Verolop:
## Description

this PR resolves component images at reconcile time instead of freezing them into the spec when a cluster is created.

Today the webhook writes the current images into `spec.images`, where they are indistinguishable from deliberate pins and never change again. A cluster stays on the images it was created with, and updating several means editing each one by hand. Now unset images come from the operator's configuration on every reconcile, and anything set in the spec stays a pin.

The defaults are also no longer fixed at build time. `MULTIGRES_IMAGE_*` on the operator Deployment overrides them per component, so a deployment can run its own images without patching `image_defaults.go`.

## Main changes

- The webhook stops writing images. The reconciler fills unset fields, existing values remain pins. Existing clusters keep their six persisted values and stay fully pinned, so nothing moves under them. There is no migration.
- `MULTIGRES_IMAGE_*` deployment-level overrides for the default set, documented in `docs/configuration.md`.
- `--image-update-strategy=lazy` holds a cluster on its current images until `spec.imageUpdatePolicy.acknowledgedRevision` names the new set, so clusters update one at a time. Default stays `immediate`.
- The Deployment uses `Recreate` so old and new webhook behavior are never served together.

## Testing

Unit tests cover default resolution, explicit and partial pins, lazy adoption, bad acknowledgements, corrupt records, and status fallback. The reconcile test checks images on child resources before and after acknowledgement. e2e is green.


niconosenzo (approved):
lgtm

PR #561, opened 2026-08-04
fix(e2e): stabilise shared runs and fix deletion fixture

Verolop:
## Description

this PR addresses two independent failures in the nightly compatibility suite rooted in our e2e tests: a deletion-test fixture whose 26-character cluster name exceeds the operator’s 25-character limit, and contention between shared E2E packages that causes the templated suite to time out waiting for a primary. It also corrects the discrepancies in the name limit docs and captures sanitised namespace diagnostics before failed-test cleanup.
## Main changes

- Runs the non-templated shared packages with bounded package concurrency, waits for their namespaces to terminate, then runs `shared/templated` alone.
- Preserves explicit test selection and the existing dedicated-mode behaviour.
- Captures namespace logs, events, and minimised Multigres resource status before cleanup, including `Shard.status.podRoles` for better signal.
- Uploads failure diagnostics for five days without retaining failed namespaces or enumerating Secret objects.
- Fixes the invalid 26-character deletion fixture and corrects the documented MultigresCluster name limit from 30 to 25 characters.

## Testing

Ran the e2e framework unit tests and vet with the `e2e` build tag, compiled the deletion e2e package, and validated the workflow with `actionlint` and YAML parsing. Also ran `git diff --check` and verified the complete shared-package selection matrix.

### DIFF
--- a/.github/workflows/_reusable-e2e.yaml
+++ b/.github/workflows/_reusable-e2e.yaml
@@ -217,10 +217,10 @@ jobs:
           esac
 
           package_file="$RUNNER_TEMP/e2e-packages"
+          : > "$package_file"
           if [ -z "$TEST_SPEC" ]; then
-            printf './test/e2e/%s/...\n' "$MODE" > "$package_file"
+            go list -tags=e2e "./test/e2e/${MODE}/..." > "$package_file"
           else
-            : > "$package_file"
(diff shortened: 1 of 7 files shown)


niconosenzo (approved):
lgtm thanks!

PR #562, opened 2026-08-04
chore(pvc): change default multigrescluster pvc retain policy

niconosenzo:
MultigresCluster now owns the volumes lifecycle policy, but the ownership is only honoured when the retain policy is set to "Delete".

- data volumes are kept (orphaned) to a max set by pvcOrphanReplicasThreshold
- topo volumes are deleted with the multigresCluster CR (note: they survive to a toposerver statefulset removal)


Verolop (approved):
yay! 🚀

PR #564, opened 2026-08-06
fix(cert): add per-pool wildcard SANs to pgbackrest server cert

niconosenzo:
pgbackrest clients connect to a replica pod's per pooler, per-cell headless-Service FQDN (<pod>.<headless-svc>.<ns>.svc.cluster.local) for pg2 backups, but the server cert only had the generic pgbackrest.<ns>.svc[.cluster.local] SANs, causing TLS hostname verification to fail with:

```
[CryptoError] unable to find hostname
'...' in certificate common name or subject alternative names.
```

Retrieve each pool/cell's headless Service and add a scoped wildcard SAN per one (a single *.<ns>.svc... wildcard can't match since the headless-svc name is an extra label). Also rotate the server cert when desired DNS SANs drift from what's issued (not just on expiry/CA change), so existing deployments self-heal.

### DIFF
--- a/pkg/cert/manager.go
+++ b/pkg/cert/manager.go
@@ -368,7 +368,7 @@ func (m *CertRotator) ensureServerCert(
 		return m.ensureServerCert(ctx, ca, depth+1)
 	}
 
-	// Check if near expiry OR if CA changed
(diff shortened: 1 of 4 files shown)


Verolop (approved):
lgtm!

PR #567, opened 2026-08-10
feat(gateway): expose multigateway failover request buffering, default-on

haritabh17:
Exposes the multigateway failover request buffer through the operator CRDs (MUL-1286), so planned failovers buffer client requests instead of surfacing errors.

## What

Adds a typed `buffer` block to the multigateway spec at every layer (CellTemplate, cluster cell overrides, inline cell spec), merged per-field by the resolver:

| CRD field | multigateway flag | binary default |
|---|---|---|
| `buffer.enabled` | `--buffer-enabled` | `false` (operator sets `true`) |
| `buffer.window` | `--buffer-window` | `10s` |
| `buffer.maxFailoverDuration` | `--buffer-max-failover-duration` | `20s` |
| `buffer.minTimeBetweenFailovers` | `--buffer-min-time-between-failovers` | `1m` |
| `buffer.size` | `--buffer-size` | `1000` |
| `buffer.drainConcurrency` | `--buffer-drain-concurrency` | `1` |

Unset fields emit no flag, leaving the binary defaults in effect. `enabled` is always emitted explicitly (`--buffer-enabled=true|false`) so an operator-level opt-out survives any future flip of the binary default.

## Default-on rationale

The binary ships with buffering off, but the operator provisions production-shaped HA clusters where transparent planned failover is the expected behavior — so the opinionated distribution defaults `enabled: true` while the binary stays neutral. Opt out per cell (or per template) with `buffer: {enabled: false}`.

**Behavior change:** existing clusters picking up this operator version start buffering during planned failovers (up to the 10s default window per request) instead of failing fast. Clients that prefer immediate errors should set `enabled: false`.

## Validation

The cluster webhook dry-run-resolves each cell (template + overrides + inline, same path as the reconciler) and validates the resolved buffer with the same rules as the binary's `Config.Validate`:

- `window` / `maxFailoverDuration` must be positive; `minTimeBetweenFailovers` must be non-negative (`0s` = no enforced gap, valid to the binary).
- `window <= maxFailoverDuration`, applying the binary defaults for unset fields — so setting only `window: 30s` is rejected at admission instead of crashlooping every gateway at startup.
- Checks are skipped when buffering is explicitly disabled, mirroring the binary.
- CellTemplate create/update is now validated too (webhook gained `create;update` verbs), so a template write can't bypass admission and ship a bad config to the Deployments. Templates are judged by what their consumers produce, not in isolation: the standalone check applies intra-field rules plus the cross-field ordering only when the template sets both fields — a template with only `window: 30s` is valid, since a consuming cluster may supply `maxFailoverDuration` via overrides. On both create and update, the webhook dry-run-resolves every consuming cluster's cells against the incoming template content — including implicit consumers of the `default` fallback template, which can gain consumers before it exists — and validates the merged result (buffer rules and resource requirements). The dry-run loop is a single shared implementation (`Resolver.ValidateResolvedGateways`) used by cluster admission and the template webhook, and a drift-guard test pins the operator's copy of the binary buffer defaults to the imported multigres module's values. Deleting a CellTemplate is also guarded: explicit references block deletion outright (no reconcile-time label pre-filter, so not-yet-reconciled clusters count), and deleting the `default` fallback dry-runs the post-deletion state (empty template) for implicit consumers, rejecting only if their resolved configs would break. The three duration fields additionally carry a Go-duration CRD schema pattern (both `µs` and `μs` accepted), so non-duration strings are rejected even when webhooks are not serving, and a CellTemplate's own resource block (requests vs limits) is validated at template write time.

Template admission blames only the write that breaks things: metadata-only updates (annotations, finalizer removal) skip consumer validation entirely, clusters being torn down never gate template writes, and a consumer that was already invalid with the stored template content (legacy misconfiguration) does not block template edits that leave it no worse — only writes that turn a valid consumer invalid are rejected.

## Compatibility

**Gateway rolling restart on upgrade:** because buffering defaults on, every multigateway Deployment gains a `--buffer-enabled=true` arg on the first reconcile after the operator upgrade, so all gateway pods roll once (standard graceful rolling update) with no CR change.

Requires a multigres image that understands the `--buffer-*` flags (the version this operator pins: `sha-7482a35` or later). Clusters pinning `spec.images.multigateway` to an older image must upgrade the image alongside the operator — opting out via `buffer: {enabled: false}` does not avoid this, because `--buffer-enabled=false` itself would be an unknown flag to an old binary. This matches the operator's standing contract that images track its pinned multigres version.

## Sizing note (measured)

On a live cluster, gateways typically perceive a planned failover for ~8.3s (consensus + topology propagation) with a longer tail, so the binary's 10s default window can still fail a request or two per failover. `window: 20s` masked all planned failovers in testing; the samples README documents this guidance.

## Verification

- `make check` and `make test-integration` green at every layer (resolver merge/defaulting, webhook admission, cell controller flag emission, cluster-handler resolution/enforcement).
- Live-verified on a kind dev cluster and on a k3s cluster: watched the buffer absorb a real planned failover end-to-end — drain completed with a query buffered for 6.15s instead of erroring.

## Known limitation

`minTimeBetweenFailovers` is expressible through the CRD but its effect is per-gateway-process state; a rolling restart of gateways resets the enforced gap. This matches the binary's behavior and is not operator-addressable.


pkg/resolver/cell.go
```diff
@@ -127,3 +135,53 @@ func mergeCellConfig(
...
+
+// defaultGatewayBuffer turns failover buffering on unless the user disabled it
+// explicitly. All other buffer fields stay unset so the multigateway binary's
+// own defaults apply.
+func defaultGatewayBuffer(spec *multigresv1alpha1.MultigatewaySpec) {
+	if spec.Buffer == nil {
+		spec.Buffer = &multigresv1alpha1.GatewayBufferConfig{}
+	}
+	if spec.Buffer.Enabled == nil {
+		spec.Buffer.Enabled = ptr.To(true)
```
niconosenzo, line 185: Just a note, this will trigger a rollout on every multigateway pod on legacy clusters, if that's what you were expecting then all good.

niconosenzo (approved):
lgtm! left a nit comment.

pkg/resolver/cell.go
haritabh17: Yupp, want it to get enabled on all clusters by default.

PR #568, opened 2026-08-11
feat(shard): remediate quarantined poolers by wiping and recreating

mkindahl:
## What & why

This is **Layer 2** of the unrecoverable-postgres work (Layer 1 is the pooler-side self-quarantine in multigres/multigres). When a multipooler's PostgreSQL is *unrecoverably* failing to start — the classic case being a genuinely diverged standby — the pooler self-quarantines: it writes `LIFECYCLE_QUARANTINED` to its own topology record and marks itself cohort-`INELIGIBLE`.

The problem this PR solves — and **why a rebuild is necessary**: a quarantined pod is invisible to Kubernetes. postgres is a child of pgctld, so both containers stay `Running` and the pod stays `Ready` (readiness gates on the gRPC control plane, not postgres).

Crucially, **the cluster does not automatically add a replacement pooler.** `createMissingResources` reconciles each replica index `0..replicas-1` and only creates a pod/PVC for an index that has **no** pod. The quarantined pod is still `Running` and still occupies its index, so it is **counted as a present replica** — the index isn't missing, so `createMissingResources` creates nothing. The shard simply sits one healthy replica short indefinitely, with a zombie holding the slot. Nothing else evicts it either: pooler pruning only fires when a pod is *absent*, and Kubernetes sees `Ready`. So no fresh standby appears on its own.

That is exactly why the quarantined pooler must be **rebuilt**. And a plain pod restart on the same PVC just FATAL-loops again — the on-disk WAL is the problem — so the rebuild has to discard the bad data and re-bootstrap a fresh standby from backup. That is what this PR does.

## What it does

The shard controller now remediates a quarantined pooler by **wiping it in place**: delete the backing pod **and hard-delete its data PVC**, so the existing `createMissingResources` recreates both at the same identity/index and the fresh, empty volume makes pgctld restore from backup and rejoin as a standby.

- **`topo.GetQuarantinedPods`** surfaces the precise per-pod quarantine signal — pod name plus the human-readable `reason` recorded on the topology lifecycle entry (e.g. "postgres failed to recover for 5m0s across 60 attempts …").
- **`GetPoolerStatus`** now maps a quarantined pooler to a distinct **`QUARANTINED`** role instead of `DRAINED`. It's visible in `Shard.Status.PodRoles`, excluded from the scale-down health gate, and — deliberately — no longer drives the stand-in-replica path (which keyed on `DRAINED`). So a quarantined pooler is replaced in place rather than having a stand-in provisioned at a new index while the bad pod lingers.
- **`reconcileQuarantineRemediation`** runs in the topo-aware data-plane phase, before the drain state machine.

## Why rebuild in place, instead of creating a new server and setting the quarantined one aside?

That "provision a fresh replica now, keep the bad one for later inspection" model is essentially what the code did *before* this PR: quarantined mapped to the `DRAINED` role, which **bumped** the effective replica count (`effectiveReplicas = replicas + drainedCount`) so a **stand-in** was created at a new, higher index while the quarantined pod kept running. We deliberately moved away from it:

- **Fixed per-index identity.** Pods and their data PVCs are per-index (StatefulSet-like: `…-0`, `…-1`, `data-…-1`). You can't bring up a second pod at the *same* identity, so "create a new one immediately" means a *different* index — leaving the replica set non-contiguous, and (since nothing removed the quarantined pod) permanently at N+1 pods, one of them a dead zombie pinning a full-size data PVC indefinitely.
- **No availability to preserve.** A quarantined replica isn't serving anyway — postgres is down and it's cohort-`INELIGIBLE`/dropped from consensus. So keeping it "to the side" buys zero availability; it's pure cost (a wasted pod slot + volume). Rebuilding the same index converges to exactly the desired replica set with no leftover state to later reconcile away.
- **Forensics has a better tool.** The failure cause is already captured in the topology lifecycle `reason`. If the diverged bytes themselves are needed, a `VolumeSnapshot` taken before the wipe preserves them without pinning a live pod/PVC — a follow-up could add that. A permanently-retained zombie pod is a poor forensics mechanism.

## Gating (all required before any destructive wipe)

1. The pooler **self-quarantined** in topology (operator never initiates it).
2. A **healthy backup exists** (`backuphealth.ConditionHealthy`) — the data-safety gate. The hard-delete is irreversible, so with no restore source we must not destroy the last copy. Absent/false blocks remediation and emits `QuarantineRemediationBlocked`.
3. The pod is **not the primary** (a quarantined node's postgres is down so it shouldn't be primary; guarded defensively — primary loss is a failover concern).
4. The pod is **old enough** (≥ 3m) — a stale-record guard so a freshly recreated pod that's still (re)bootstrapping isn't wiped on a leftover `QUARANTINED` record.
5. The **rest of the pool is otherwise healthy** — don't pile destructive actions on during a broader outage.
6. **One pod per reconcile**, then requeue.

The reason is echoed on the `QuarantineRemediation` event so `kubectl describe shard` shows the postgres-level cause.

## Deliberately out of scope

- **No `UNRECOVERABLE` lifecycle state / no quarantine attempt-cap.** A reconstructed pod is just a new standby, so if reconstruction repeatedly fails (e.g. a corrupt backup) that is a *general* standby bring-up / restore failure — flagged by backup-health and standby-provisioning monitoring — independent of whether the pod was ever quarantined. Encoding it into the quarantine lifecycle would conflate two orthogonal concerns.
- Note the backup-health gate only detects the coarse cases (no / incomplete / stale backup from pgBackRest metadata); a complete-but-corrupt backup is caught at restore time as that general standby failure, not here.

## Testing

- Unit: `reconcile_quarantine_internal_test.go` — wipes-when-healthy (asserts pod + data PVC deleted and the reason on the event), defer-when-too-young, defer-when-pool-unhealthy, **defer-when-no-healthy-backup** (never wipe the last copy), and no-op-when-nothing-quarantined. Plus the topo role test updated to expect the `QUARANTINED` role.
- Full operator unit suite green; `golangci-lint` clean.

## Follow-ups before merge (need envtest, not the fake client)

- **PVC finalizer + same-name recreate convergence:** the fresh PVC reuses the deterministic name, so it can only be created once the old one fully terminates (pod gone → `pvc-protection` finalizer clears → PV reclaimed). The fake-client tests don't model finalizers; this needs the `//go:build integration` envtest path.
- **PV reclaim policy:** a true wipe depends on `Delete` (dynamic-provisioning default); on a `Retain` storage class the volume is only released and the data wouldn't actually be reclaimed. Worth asserting/handling explicitly.
- The now-dormant `DRAINED` stand-in machinery (unused once quarantine no longer maps to `DRAINED`) can be removed in a follow-up.

## Related PRs

- **Layer 1 (pooler):** multigres/multigres#1365 — the producer that sets the `LIFECYCLE_QUARANTINED` signal this PR consumes.
- **Follow-up:** multigres/multigres-operator#570 — removes the now-dormant `DRAINED` stand-in machinery noted above (stacked on this PR).

Part of MUL-1009.

### DIFF
--- a/pkg/data-handler/topo/pooler.go
+++ b/pkg/data-handler/topo/pooler.go
@@ -22,7 +22,10 @@ const deadPoolerReason = "operator: no backing pod for pooler"
 // PoolerStatusResult holds the result of querying pooler roles from the topology.
 type PoolerStatusResult struct {
 	// Roles maps hostname to its operator-facing role
-	// (PRIMARY, REPLICA, DRAINED).
+	// (PRIMARY, REPLICA, QUARANTINED). Shutdown poolers are omitted. QUARANTINED
+	// poolers are surfaced for visibility but are not routed and do not drive the
+	// stand-in-replica path; they are replaced via quarantine remediation (see
+	// GetQuarantinedPods).
 	Roles map[string]string
 	// QuerySuccess indicates whether all topology queries succeeded.
 	QuerySuccess bool
@@ -52,10 +55,18 @@ func GetPoolerStatus(
 				if isLifecycleShutdown(p.Multipooler) {
 					continue
 				}
+				// Quarantined poolers are unrecoverable (postgres cannot start).
+				// They are surfaced with a distinct QUARANTINED role so they are
+				// visible in Shard.Status.PodRoles, but they are not routed and do
+				// not drive the stand-in-replica path (which keyed on DRAINED).
+				// The operator replaces them via quarantine remediation (delete pod
+				// + wipe data PVC + re-bootstrap from backup); GetQuarantinedPods
+				// carries the reason for that.
 				roleName := "REPLICA"
-				if isLifecycleQuarantined(p.Multipooler) {
-					roleName = "DRAINED"
-				} else if IsPrimaryPooler(p.Multipooler) {
+				switch {
+				case isLifecycleQuarantined(p.Multipooler):
+					roleName = "QUARANTINED"
+				case IsPrimaryPooler(p.Multipooler):
 					roleName = "PRIMARY"
 				}
 				// Match the topology entry to an actual managed pod.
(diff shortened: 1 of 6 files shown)


niconosenzo (approved):
lgtm! Thanks!

PR #570, opened 2026-08-12
refactor(shard): remove dormant DRAINED stand-in machinery

mkindahl:
## Summary

Follow-up cleanup that removes the now-dormant **DRAINED stand-in** machinery. Quarantine remediation (#568) supersedes it: a pooler whose PostgreSQL is unrecoverable is now marked `QUARANTINED` and the operator wipes + re-bootstraps the pod in place, rather than keeping the bad pod alive and provisioning a stand-in replica at a higher index. With that path in place, nothing ever sets the `DRAINED` topology role anymore, so all the code that reacted to it is dead.

## What's removed

- `countDrainedPods` and the `effectiveReplicas = replicas + drainedCount` computation. `createMissingResources` and `handleScaleDown` now operate on the user-desired replica count directly.
- `syncDrainedLabels` and the `multigres.com/role` (`LabelPodRole`) label it maintained, plus the `DRAINED` branch in `cleanupDrainedPod` that unconditionally orphaned a pod's PVC.
- The `DRAINED` short-circuits in `isDrainStale` and `isPoolHealthy`. `isPoolHealthy` still excludes `QUARANTINED` pods (which are expected to be unhealthy while remediation replaces them).

## No behavior change for live paths

Rolling updates, scale-down, and quarantine remediation are all unaffected — `DRAINED` was never produced once quarantine landed. Tests that exercised the removed stand-in behavior are deleted (`TestDrainedPodReplacement`, the `syncDrainedLabels` error case) or refolded onto the index-vs-replicas semantics that remain (`TestCleanupDrainedPod_PVCDeletion`, `TestHandleScaleDown_ConcurrentDrainPrevention`, the `isPoolHealthy`/`isDrainStale` cases).

Part of MUL-1009.

### DIFF
--- a/pkg/data-handler/topo/pooler.go
+++ b/pkg/data-handler/topo/pooler.go
@@ -57,8 +57,7 @@ func GetPoolerStatus(
 				}
 				// Quarantined poolers are unrecoverable (postgres cannot start).
 				// They are surfaced with a distinct QUARANTINED role so they are
-				// visible in Shard.Status.PodRoles, but they are not routed and do
-				// not drive the stand-in-replica path (which keyed on DRAINED).
+				// visible in Shard.Status.PodRoles, but they are not routed.
 				// The operator replaces them via quarantine remediation (delete pod
 				// + wipe data PVC + re-bootstrap from backup); GetQuarantinedPods
 				// carries the reason for that.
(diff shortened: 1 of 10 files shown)


Verolop (approved):
lgtm!
Remaining references to DRAINED() need to be addressed; can be included here or as a follow up

PR #572, opened 2026-08-14
feat(postgres-config): reload GUC changes in place instead of restarting

GuptaManan100:
## Summary

Apply reload-safe `postgresql.conf` changes to running PostgreSQL in place via a configuration reload (SIGHUP) instead of recreating pods, so a change confined to reload-safe parameters converges without a rolling restart or dropped connections.

- Classify each PostgreSQL GUC as restart- vs reload-required — a **static** check in the operator (from the `pg_settings` `context`), not something the running server reports — and split the single rendered-config hash into a restart-hash and a reload-hash, so a reload-safe change no longer moves the pod spec-hash.
- Apply reload-safe changes via the multipooler `ReloadConfig` RPC. The reload is **verified, not timed**: the operator passes the exact settings it expects and the pooler reloads only once the mounted file actually carries them, so a not-yet-synced ConfigMap is retried rather than falsely marked applied.
- Report both hashes in `status.postgresConfig`, so `inProgress` stays true through an in-place reload just as it does through a rolling restart.
- Bump the `github.com/multigres/multigres` module and the data-plane image defaults to the revision that ships the verifying `ReloadConfig` RPC.

## What's implemented

- **Catalog + classifier** (`pkg/postgresconfig/classify.go`): regenerated the PG17 catalog with a `context` column and added `RequiresRestart` — postmaster/internal ⇒ restart; every other context ⇒ reload; unknown/namespaced (e.g. `cron.*`) ⇒ restart (conservative: needlessly restarting is disruptive but correct, reloading a restart-only param would silently fail to apply). Regenerating also recovered 7 GUCs the previous generator dropped (multi-line `guc_tables.c` entries), fixing their spurious "unknown parameter" admission rejection.
- **Hash split** (`pkg/postgresconfig/hash.go`): `SplitConfig` parses the rendered config into effective settings and, in a single pass, partitions them by `RequiresRestart` into a restart-hash, a reload-hash, and the reload-safe name→value map. Only the restart-hash feeds `ComputeSpecHash`, so a reload-only change leaves the spec-hash untouched.
- **Reload reconcile step** (`reconcileReloadState`): for each pool pod whose reload-hash annotation drifts from the shard's desired reload-hash, it calls `ReloadConfig` with the expected reload-safe settings and only stamps the pod current when the pooler confirms the reload took effect against a file that already carries them. Reloads are unordered (non-disruptive); draining, terminating, and restart-pending pods are skipped so a pending recreation applies the config instead. A confirmed reload emits `ConfigReloaded`; a reload-classified param the server reports needs a restart emits a `ConfigReloadNeedsRestart` warning without stamping.
- **Status** (`status.go`): `postgresConfig.inProgress` requires both the restart-hash and the reload-hash on every live pod.
- **Docs** (`docs/postgresql-configuration.md`): new "Applying a change: reload vs. restart" section — the two classes with their contexts and examples, that the classification is static and operator-owned, which path runs for a given edit, and the verify-not-time reload with its events.
- Unit + envtest coverage: classifier, hash-split determinism, the reload state machine over a fake RPC (verified/stamped, not-synced/retry, needs-restart, and skip cases), and a reload-vs-restart rollout integration test.

## Verified end-to-end

The reload path is validated against a live cluster (kind/OrbStack), not just envtest. In the e2e suite, a reload-safe `work_mem` change lands in the running server via SIGHUP with **pool pod UIDs unchanged** (proving no restart), while a `max_connections` (postmaster-context) change still drives the primary-last rolling restart. Both effective values are read back with `SHOW` through the gateway.


pkg/postgresconfig/hash_test.go
mkindahl, line 1: Would be good to have a test that cover more subtle cases:

Starting:
```conf
work_mem = '4MB'
```

Should not trigger a reload:
```conf
work_mem = 4MB
```

Should not trigger a reload:
```conf
work_mem = '4 MB'
```

Should trigger a reload, but will error (how is that handled?):
```conf
work_mem = 4mb
```

Same for a restart-parameter such as `shared_mem`.

pkg/postgresconfig/hash_test.go
```diff
@@ -0,0 +1,150 @@
...
+		in, want string
+	}{
+		{"'128MB'		# (change requires restart)", "'128MB'\t\t"},
+		{"'HIGH:MEDIUM:+3DES:!aNULL' # allowed SSL ciphers", "'HIGH:MEDIUM:+3DES:!aNULL' "},
+		{"on", "on"},
+		{"'%h %m [%p] %q%u@%d '", "'%h %m [%p] %q%u@%d '"},
+		// '#' inside a single-quoted value is preserved.
+		{"'a#b'", "'a#b'"},
+		// escaped '' inside a quoted value does not end the string.
+		{"'it''s #1' # trailer", "'it''s #1' "},
```
mkindahl, line 115: Missing error cases that could cause strange problems, such as:

```
'4MB
'4MB # comment and no trailing quote
'4MB' # comment with '' inside
'just a test ''
```

Probably some more weird combinations are possible.

pkg/postgresconfig/validate_test.go
```diff
@@ -108,7 +108,7 @@ func TestValidate_AggregatesAllProblems(t *testing.T) {
 func TestCatalogLoaded(t *testing.T) {
 	// The embedded catalog must be non-trivial and contain well-known params.
 	if len(catalog) < 300 {
-		t.Errorf("catalog has %d entries, expected the full PG17 set (~382)", len(catalog))
+		t.Errorf("catalog has %d entries, expected the full PG17 set (389)", len(catalog))
```
mkindahl, line 111: This will keep changing
```suggestion
		t.Errorf("catalog has %d entries, expected the full PG17 set", len(catalog))
```

niconosenzo (approved):

mkindahl (approved):
Looks good. Some minor test additions to cover corner-cases.

PR #581, opened 2026-08-17
feat(shard): add custom queries to the postgres_exporter sidecar

niconosenzo:
Serve a custom queries file via --extend.query-path, alongside the default collectors. Adds instance-level metrics commonly needed by monitoring dashboards: database disk size, WAL disk size, physical replication lag, total backends, per-role connection breakdown, and the configured max_connections cap.

The queries file is embedded in the operator, delivered as a per-shard ConfigMap (server-side applied, owner-ref'd), and mounted read-only into the exporter sidecar of every pool pod.

Updates on the ConfigMap won't trigger a rollout, this is a design decision to avoid unwanted pooler pod restarts.


Verolop (approved):
🚀

PR #590, opened 2026-08-18
fix(drain): leave cohort removal to multiorch

Verolop:
## Description

this PR removes the operator's direct cohort and topology mutations from the drain path.

A drain now only advances the pod through the drain states and permits Kubernetes termination. The pooler shutdown path and Multiorch reconcile cohort membership. A pooler that is already missing continues to use the existing dead-pooler path, which writes `LIFECYCLE_SHUTDOWN` instead of deleting its topology record.

This avoids the operator racing orch by removing a member that Multiorch may otherwise re-observe and re-add.

## Main Changes

- Remove the direct `UpdateConsensusRule(COHORT_REMOVE)` call.
- Remove forced Multipooler unregistration and its drain-only helpers.
- Remove the drain RPC/topo dependencies and the obsolete test scaffolding around them.
- Keep the drain state machine focused on pod state transitions and deletion readiness.
- Add Kind coverage for:
  - external replica deletion and replacement;
  - external primary deletion, replacement, and failover;
  - scale-down from three poolers to two while the shard remains healthy.

## Verification

Kind run completed successfully against the locally built operator image.


pkg/data-handler/drain/drain.go
```diff
@@ -75,225 +60,35 @@ func ExecuteDrainStateMachine(
...
-					}
-				}
-			}
-		}
 		return UpdateDrainState(ctx, k8sClient, pod, metadata.DrainStateAcknowledged)
 
 	case metadata.DrainStateAcknowledged:
-		if myPooler != nil {
-			cellName := pod.Labels[metadata.LabelMultigresCell]
-			if err := topo.ForceUnregisterPod(ctx, store, shard, pod.Name, cellName); err != nil {
```
dweitzman, line 241: Just to confirm, I'm guessing something elsewhere is making sure the pod transitions to SHUTDOWN lifecycle stage before the operator stops acting (plus associated tests)?

test/e2e/shared/drain/drain_test.go
```diff
@@ -0,0 +1,261 @@
...
+			switch freshShard.Status.PodRoles[pod.Name] {
+			case "PRIMARY":
+				primary = pod.DeepCopy()
+			case "REPLICA":
+				if replica == nil {
+					replica = pod.DeepCopy()
+				}
+			}
+		}
+		return primary != nil && replica != nil, nil
```
dweitzman, line 205: Do these tests run with real orch, etc.? If so, could potentially assert here than the cohort has removed the shut down pooler

If this doesn't run with real multipooler & orch, please ignore :)

dweitzman (approved):

pkg/data-handler/drain/drain.go
Verolop: Yes. The operator no longer has a direct cohort-removal action. For a live scale-down, Multipooler persists `LIFECYCLE_SHUTDOWN` as part of its shutdown; the Kind E2E now asserts that state. For the missing/dead fallback, `MarkDeadPoolers` writes `LIFECYCLE_SHUTDOWN` without deleting the topology entry or changing cohort membership; that is covered by `TestMarkDeadPoolers`.

test/e2e/shared/drain/drain_test.go
Verolop: Yes, it runs real Multiorch and Multipooler in Kind. I extended the scale-down E2E to wait for the removed pooler’s topology entry to reach `LIFECYCLE_SHUTDOWN`, then read the surviving primary’s decided cohort through Multiadmin and assert that it no longer contains the removed pooler.

PR #593, opened 2026-08-19
fix(exporter): disable settings metrics to unblock custom queries

niconosenzo:
A GUC with a NULL short_desc in pg_settings crashes postgres_exporter's internal settings probe on every scrape ("Scan error on column index 3, name \"short_desc\""). That error aborts the scrape before --extend.query-path queries run.

Current error log:
```
level=ERROR source=postgres_exporter.go:684 msg="error scraping dsn" err="error retrieving settings: Error retrieving rows on \"localhost:5432\": pg sql: Scan error on column index 3, name \"short_desc\": converting NULL to string is unsupported" 
[...]
```


niconosenzo:
Fixed on https://github.com/prometheus-community/postgres_exporter/pull/1327, proper fix is to bump the postgres exporter version.

PR #598, opened 2026-08-20
fix(shard): share filesystem backup PVC across cells

### LINKED ISSUE #597: Bootstrap multi-cell filesystem shards from a single backup repository
The operator creates a filesystem backup PVC per cell. This is incompatible with the bootstrap protocol for a shard that has poolers in more than one cell.

A pooler initializes `multigres.current_rule` before acquiring the shard backup lease. The initial `0.1` rule includes the bootstrap pooler identity and its creation timestamp. The lease holder then writes that database state into the first backup.

With one repository per cell, each cell can create its own first backup. In the failing E2E run:

```text
az1: p-3ec3fdea created 20260820-172931F
az2: p-a04d15ac created 20260820-172937F
```

Poolers restored from those separate backups and reported different `0.1` rule contents. Multiorch compares complete rule positions during pre-vote and correctly refused to elect a primary:

```text
recruits az1_… and az2_… agree on rule number but disagree on content;
inconsistent discovery result
```

The backup lease is shard-scoped. The filesystem repository must be shard-scoped as well.

Change the operator to create one RWX filesystem backup PVC per shard and mount it in pooler pods in every cell. The PVC name and reconciliation must not include the cell name.

Add an E2E case for a multi-cell shard using filesystem backup storage. It should verify that all poolers mount the same backup claim and that bootstrap elects a primary.

Verolop:
## Description

this PR makes filesystem backup repositories shard-scoped, matching the scope of the backup lease.

During bootstrap, a pooler writes its initial `multigres.current_rule` before taking the shard backup lease. With one filesystem repository per cell, separate cells can preserve different first backups. When those poolers later bootstrap together, Multiorch sees different `0.1` rule positions during pre-vote and refuses to elect a primary.

A shard needs one filesystem repository, visible to every pooler that might restore from it.

## Main changes

- Remove the cell from filesystem backup PVC naming and reconciliation, leaving one claim per shard.
- Mount that claim in pooler pods in every cell, so every bootstrap candidate restores from the same repository.
- Warn when a multi-cell filesystem shard uses RWO storage. The shared claim needs RWX storage across cells; otherwise S3 remains the supported shared repository.
- Add a multi-cell filesystem e2e case that verifies the shared claim and primary election.
- Update the backup docs to describe the shard-wide repository rather than the old cell-local layout.

## Testing

Ran the resolver and shard-controller test suites, compiled the verification e2e package with the `e2e` build tag.

fixes https://github.com/multigres/multigres-operator/issues/597


pkg/resolver/validation.go
```diff
@@ -561,21 +561,23 @@ func (r *Resolver) ValidateClusterLogic(
...
-								))
-							}
+							poolerMounts += int64(len(pool.Cells)) * int64(replicas)
+						}
+
+						if poolerMounts > 1 {
+							warnings = append(warnings, fmt.Sprintf(
+								"Shard '%s' uses filesystem backups with ReadWriteOnce (RWO) storage but %d poolers mount the shard-wide backup PVC. "+
+									"This configuration may fail when poolers are scheduled on different nodes. "+
+									"Configure ReadWriteMany (RWX) storage, ensure node affinity, or use S3 backups.",
```
niconosenzo, line 577: Should this configuration be rejected instead of returned as an admission warning?

niconosenzo (approved):
Solid, left a non-blocker question.

test/e2e/shared/verification/verification_test.go
```diff
@@ -4,21 +4,198 @@ package verification_test
...
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/utils/ptr"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	multigresv1alpha1 "github.com/multigres/multigres-operator/api/v1alpha1"
+	"github.com/multigres/multigres-operator/pkg/util/metadata"
 	"github.com/multigres/multigres-operator/test/e2e/framework"
 )
 
+const backupVolumeName = "backup-data"
```
cuongdo: nit: this duplicates `BackupVolumeName` from `pkg/resource-handler/controller/shard/containers.go`

cuongdo (approved):
Approved with one tiny nit. Feel free to disregard.

pkg/resolver/validation.go
Verolop: Yeah, fair question. I left this as a warning because the current defaults can already produce this setup: filesystem is the default backup backend, an unset access mode becomes RWO, and a resolved shard can have more than one pooler. Making it an admission error here would start rejecting existing default configurations, even though their input has not changed.

The problem is specifically when those poolers land on different nodes. RWO can technically be mounted read-write by multiple pods on one node. We do not recommend that as the multi-pooler backup setup, RWX or S3 is the intended path, but making this an error needs a separate change to the filesystem defaults rather than just changing this validation.

PR #601, opened 2026-08-21
fix(topology): make shared etcd roots cluster-scoped

Verolop:
## Description

this PR makes it safe for multiple Multigres clusters to use the same etcd instance.

Topology roots already existed, but their defaults did not include the cluster identity. Every cluster used `/multigres/global` for global topology and `/multigres/<cell>` for local topology. A Cell without its own topology server also pointed back at the global root. As soon as two clusters shared an etcd instance, they could read and write the same keys.

Default roots now look like this:

```text
/multigres/<cluster-id>/global
/multigres/<cluster-id>/<cell>
```

The project-reference annotation is used as the cluster ID when available. Otherwise, the namespace and MultigresCluster name are used together. Global and cell topology still use the same etcd address; the roots give each cluster and cell its own keyspace.

## Main changes

- Add a shared helper for deriving global and cell roots so resolution and Cell registration use the same layout.
- Default managed and external topology to cluster-scoped roots while leaving explicit roots unchanged.
- Register Cells against the shared etcd address with their own cell root, including Cells without a separate local topology server.
- Pass the resolved topology address and root to MultiAdmin instead of hardcoding `/multigres/global`.
- Update the observer to resolve the root from the cluster and its CoreTemplate rather than checking the old hardcoded path.
- Reject `global` as a Cell name because the topology client reserves it for the global connection.
- Add coverage for two clusters with the same Cell name sharing one backing store without seeing each other’s topology.

## Testing

Ran the complete operator test suite and the complete observer test suite separately. Regenerated the CRDs.


rafael (approved):
Good catch. LGTM

PR #603, opened 2026-08-24
fix(shard): detect postgres posture inconsistencies

niconosenzo:
Cross-check Shard topology roles against each multipooler’s observed postgres posture during normal reconciliation.

- Query `MultipoolerClient.Status()` and expose posture in `status.podPostures`
- Detect multiple observed primaries and topology/posture role mismatches
- Require two consecutive inconsistent observations before degrading
- Treat partial scans and unknown postures as incomplete evidence
- Prevent incomplete observations from clearing a confirmed fault
- Atomically publish the posture condition and derived Shard phase
- Emit warning events and expose a shard-level inconsistency metric
- Updated the shard API, generated deepcopy, and CRD schema

A pooler advertised as a replica while postgres reports primary is surfaced as a role mismatch. Multiple observed primaries cause the Shard to become `Degraded` after debounce.


config/crd/bases/multigres.com_multigresclusters.yaml
```diff
@@ -87,7 +87,7 @@ spec:
                       storage:
                         description: |-
                           Storage defines the PVC configuration for the backup volume.
-                          This volume is shared by all pools in the shard (per-cell).
+                          This volume is shared by all pools in the shard.
```
niconosenzo, line 90: This comes from https://github.com/multigres/multigres-operator/pull/598/changes 😄

pkg/resource-handler/controller/shard/reconcile_data_plane.go
```diff
@@ -48,6 +50,21 @@ func (r *ShardReconciler) reconcileDataPlane(
...
+			childSpan.End()
+			return ctrl.Result{}, err
+		}
+		childSpan.End()
+
+		// Publish posture, condition, and phase in one status update.
+		if err := r.updateStatus(ctx, shard, rendered); err != nil {
+			return ctrl.Result{}, fmt.Errorf("update status after posture observation: %w", err)
+		}
+	}
```
Verolop, line 69: Since a direct sql change doesn’t trigger a kubernetes reconcile, we might never run the second check required by the debounce. That could leave the Shard reporting Healthy indefinitely. Maybe we could requeue after a few seconds when the first strike is recorded?

niconosenzo: Added a 5s requeue posture reconcile, it will mark the shard as degraded after failing the second strike.

pkg/data-handler/posture/posture.go
```diff
@@ -0,0 +1,208 @@
...
+			}
+			return nil, fmt.Errorf("listing poolers in cell %q for posture check: %w", cell, err)
+		}
+
+		for _, p := range poolers {
+			if isLifecycleShutdown(p.Multipooler) {
+				continue
+			}
+			podName := matchPod(p, managedPodNames)
+			if podName == "" {
```
Verolop, line 68: non-blocking: when an active topo pooler has no matching Kubernetes pod, Evaluate skips it without marking the result incomplete. That can leave PostureConsistent=True even though not all active poolers were observed. It should be rare, though.

Verolop (approved):
lgtm! left non-blocking comment

pkg/data-handler/posture/posture.go
niconosenzo: Good catch. I’ll mark the observation incomplete when an active topology pooler has no matching managed pod. Shutdown poolers will still be ignored, and incomplete evidence won’t clear an existing failure. I’ll add a regression test for this case as well.

PR #605, opened 2026-08-24
feat(topo): pass cell metadata through to the topology record

soedirgo:
Add CellConfig.Metadata and CellSpec.Metadata — an opaque string the worker writes via the MultigresCluster CR. The operator copies it verbatim into the topo Cell record without parsing or interpreting it; multiadmin then serves it to administrative clients.

TODO:
- [x] replace the soedirgo/multigres fork pin once https://github.com/multigres/multigres/pull/1404 lands

Part of MUL-1371


kangmingtay (approved):
looks good as long as we update the image_defaults.go file before merging

PR #606, opened 2026-08-24
feat(shard): support PostgreSQL init secrets

niconosenzo:
## Summary

- Add `postgresInitSecretsRef` to `MultigresCluster`
- Propagate the reference through TableGroups and Shards
- Validate the referenced Secret and JSON payload
- Mount the Secret read-only in pgctld containers
- Configure pgctld through `POSTGRES_INIT_SECRETS_FILE`
- Document configuration and lifecycle behavior
- Add unit and integration coverage

## Notes

Role entries must reference roles created by init SQL before pgctld applies the init-secrets payload. The configuration affects only newly initialized data directories.


Verolop (approved):
lgtm!

pkg/resource-handler/controller/shard/reconcile_shared_infra.go
```diff
@@ -125,6 +126,63 @@ func (r *ShardReconciler) postgresPasswordSecret(
...
+		Roles            map[string]string            `json:"roles"`
+		DatabaseSettings map[string]map[string]string `json:"database_settings"`
+	}
+	if err := json.Unmarshal(data, &payload); err != nil {
+		return fmt.Errorf(
+			"key %q in postgres init-secrets Secret %q is not valid JSON",
+			secretKey,
+			secretName,
+		)
+	}
```
mkindahl: Hmm... I think this will fail with this. It might be valid to require everything to be strings, but it seems little strict.

```json
{
  "roles": { "app": "some-password" },
  "database_settings": { "postgres": { "max_connections": 100 } }
}
```

pkg/resource-handler/controller/shard/integration_test.go
```diff
@@ -1204,6 +1395,108 @@ func TestReconcileDeletions(t *testing.T) {
...
+	}
+
+	setTestPostgresPasswordSecretRef(shard)
+	createTestPostgresPasswordSecret(t, ctx, k8sClient, shard.Namespace)
+	if err := k8sClient.Create(ctx, shard); err != nil {
+		t.Fatalf("Failed to create Shard: %v", err)
+	}
+
+	// Allow multiple reconciles before checking for pods.
+	time.Sleep(3 * time.Second)
```
mkindahl: You might want to use `Eventually` here instead to avoid a flaky test.

pkg/resource-handler/controller/shard/reconcile_shared_infra.go
```diff
@@ -125,6 +126,63 @@ func (r *ShardReconciler) postgresPasswordSecret(
...
+
+	var payload struct {
+		Roles            map[string]string            `json:"roles"`
+		DatabaseSettings map[string]map[string]string `json:"database_settings"`
+	}
+	if err := json.Unmarshal(data, &payload); err != nil {
```
mkindahl, line 183: Hmm, what happens with the JSON document below?

```json
null
```

I think you might want to check that `Roles` and `DatabaseSettings` was actually assigned something.

mkindahl (approved):
Some minor suggestions for improvement and hardening, but approving since they are non-critical.

pkg/resource-handler/controller/shard/reconcile_shared_infra.go
niconosenzo: This is intentional and matches pgctld’s schema as far as I understand. The database setting values are strings because they are passed to PostgreSQL as text. Numeric literal would also fail in pgctld, so it should be "max_connections": "100". 
I’ll clarify that in the documentation and improve the validation error so it doesn’t describe valid JSON with the wrong schema as invalid JSON.

niconosenzo: This is actually intentional and matches pgctld’s string-based schema. I clarified that numeric GUC values must be quoted.

pkg/resource-handler/controller/shard/reconcile_shared_infra.go
niconosenzo: Good catch. Fixed, top level null payloads are now rejected while roles-only and settings-only payloads remain supported.

PR #609, opened 2026-08-25
feat(topo): issue TLS certificates for the topology server

Verolop:
## Description

this PR issues the TLS credentials a topology server needs before it can require client certificates.

The topology server serves plain HTTP and authenticates nobody. One etcd instance can back several Multigres clusters, separated only by key path, so any client that reaches it can read and write every cluster's topology.

Two certificates are issued: a serving certificate for the topology server covering both Services the controller creates, and a per-cluster client credential whose common name is the cluster's topology root, so the identity a cluster proves and the keyspace it is entitled to are the same string. Both are signed by the topology issuer, since they have to chain to the same CA for the server to accept the client. Every certificate that CA signs becomes a candidate topology client, so an enabled configuration has to name its issuer rather than inherit one.

Nothing starts enforcing TLS here. The etcd listener configuration, the client certificate auth flag, the volume mounts, the component flags and the health probes are unchanged, and issuance is opt-in through `topoTLS`, off by default.

## Main changes

- Add a topology TLS configuration on MultigresCluster that propagates into the managed TopoServer and requires an explicit issuer when enabled.
- Issue a serving certificate for the topology server, and remove it with its Secret when the configuration is disabled or removed.
- Issue a per-cluster client credential rooted at the cluster's topology prefix, rejecting roots over the X.509 common name limit rather than truncating an identity.
- Read the external topology CA and client certificate secret fields, previously accepted and never used, and carry both references through the global topology server reference so Cell, TableGroup and Shard receive them.
- Expose the topology key prefix separately from the cluster root, because a range opened at the bare root also reaches a sibling cluster whose identity starts with the same characters.
- Extract the cert-manager machinery into a shared package so both controllers issue, compare and prune certificates the same way.

## Testing

Ran the complete operator unit test suite and regenerated the CRDs.


pkg/resource-handler/controller/toposerver/certificate.go
```diff
@@ -0,0 +1,112 @@
...
+	certName := multigresv1alpha1.TopoServerCertName(toposerver.Name)
+	existing, err := certs.Get(ctx, r.Client, toposerver.Namespace, certName)
+	if err != nil {
+		return err
+	}
+	// Only ever touch a Certificate this TopoServer owns, so a same-named
+	// object belonging to something else is left alone.
+	if existing != nil && !certs.OwnedBy(existing, toposerver.UID) {
+		existing = nil
+	}
```
niconosenzo: Question, the comment says a same named cert owned by something else is left alone, but if topoTLS enabled we still call certs.ApplyOne with ForceOwnership. That could either fail because it already has another controller or cause us to adopt it, correct ? if so, it will then delete it and its secret if topoTLS is later disabled. Could we return a clear conflict error instead? That would be safer and make the collision visible.

niconosenzo:
Awesome! left a small question, but it lgtm overall.

niconosenzo (approved):
🚀

PR #610, opened 2026-08-26
feat: NetworkPolicy-based ingress isolation for multiadmin

niconosenzo:
## Description

Add `spec.networkPolicy` to MultigresCluster. When enabled, the operator creates NetworkPolicies restricting ingress to the multiadmin and multiadmin-web pods: same-namespace peers plus the namespaces listed in allowedIngressNamespaces (matched via the kubernetes.io/metadata.name label). Disabling the config deletes the policies.


pkg/cluster-handler/controller/multigrescluster/reconcile_global.go
```diff
@@ -37,6 +40,52 @@ func (r *MultigresClusterReconciler) reconcileGlobalComponents(
 	if err := r.reconcileMultiadminWeb(ctx, cluster, res); err != nil {
 		return err
 	}
+	if err := r.reconcileAdminNetworkPolicies(ctx, cluster); err != nil {
```
Verolop, line 43: I think this is unlikely, but very sad if it were to happen: can you add NetworkPolicy to SetupWithManager too? Otherwise if one gets deleted, we won’t know, and we won't recreate it until the cluster changes again.

Verolop (commented):
nice! Just one little request to add a missing controller watch, and it'll be ready to merge.

pkg/cluster-handler/controller/multigrescluster/reconcile_global.go
niconosenzo: Good catch 👍

Verolop (approved):
lgtm!

PR #620, opened 2026-09-02
fix: check if shard is healthy before reconciling pool

kangmingtay:
## Summary
* When the pool is reconciled (e.g. due to a change in the postgres image version), rolling updates need to consider the overall shard health before rolling the next pod. `isShardHealthy` reports whether every pool pod across the whole shard is Ready, and does not let more than 1 pool pod in the shard be unhealthy. This guards against the following cases:
  1. when `replicasPerCell = 1` across pools, it can lead to all pools being potentially reconciled all at once
  2. when `replicasPerCell > 1` across pools, it can lead to pods being cycled through in back-to-back reconciles without waiting for the previous replacement to actually become healthy, resulting in more than one replica in the same pool down at overlapping times

* We could potentially make it more efficient in the future by allowing multiple pools to be reconciled independently at once, but that extra overhead in complexity is not worth the tradeoff at this stage

### DIFF
--- a/pkg/resource-handler/controller/shard/reconcile_pool_pods.go
+++ b/pkg/resource-handler/controller/shard/reconcile_pool_pods.go
@@ -318,28 +318,39 @@ func isPodReady(pod *corev1.Pod) bool {
 	return false
 }
 
-// isPoolHealthy returns true if all non-draining, non-terminating, non-DRAINED
-// pods that will remain after scale-down are Ready. Extra pods (index >=
-// effectiveReplicas) are excluded so an unhealthy extra pod does not block its
-// own removal. DRAINED pods are excluded because they are expected to be
-// unhealthy and should not block scale-down of stand-in pods.
+// isPoolHealthy returns true if the pool/cell has at least effectiveReplicas
+// pods, and all of them — except extras (index >= effectiveReplicas) and
+// DRAINED/QUARANTINED pods — are Ready, with none draining or terminating.
+// Extra pods are excluded so an unhealthy extra pod does not block its own
+// removal. DRAINED and QUARANTINED pods are excluded because they are
+// expected to be unhealthy and should not block scale-down of stand-in pods.
+//
+// The count check matters because a pod drained all the way to deletion
+// disappears from existingPods entirely — there is nothing left for the
+// per-pod checks below to flag as unhealthy. Without it, a pool/cell sitting
+// at zero pods for a slot (deleted, replacement not yet created) would look
+// indistinguishable from a pool that was never touched.
 func isPoolHealthy(
 	existingPods map[string]*corev1.Pod,
 	effectiveReplicas int32,
 	shard *multigresv1alpha1.Shard,
 ) bool {
+	//nolint:gosec // pod count will never exceed int32 capacity
+	if int32(len(existingPods)) < effectiveReplicas {
+		return false
+	}
 	for _, pod := range existingPods {
-		if pod.Annotations[metadata.AnnotationDrainState] != "" || !pod.DeletionTimestamp.IsZero() {
-			continue
-		}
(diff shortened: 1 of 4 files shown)


pkg/resource-handler/controller/shard/reconcile_pool_pods.go
```diff
@@ -555,6 +599,14 @@ func (r *ShardReconciler) handleRollingUpdates(
 		return nil
 	}
 
+	// Refuse to start a new drain if some other pool/cell in this shard is
+	// already mid-rollout or has not yet come back Ready.
+	if healthy, err := r.isShardHealthy(ctx, shard); err != nil {
```
Verolop, line 652: I think there may still be a race here. `isShardHealthy` reads from the controller cache, while `initiateDrain` writes directly to the API server. Since we reconcile cells sequentially, cell A can start draining and cell B may still see the pre-drain cached state, consider the shard healthy, and start another drain.

Can you please track whether a rollout has already started during this shard reconcile? That would avoid relying on cache timing. The fake client updates immediately, so it won’t catch this race.

pkg/resource-handler/controller/shard/reconcile_pool_pods.go
```diff
@@ -346,6 +348,50 @@ func isPoolHealthy(
 	return true
 }
 
+// isShardHealthy reports whether every pool pod across the whole shard — not
+// just the pool/cell currently being reconciled — is Ready, with none
+// draining or terminating
+func (r *ShardReconciler) isShardHealthy(
```
niconosenzo, line 364: Question, this is getting the list of existing pods, but between deleting and draining a pod and creating its replacement, the remaining pods might show up as healthy (and we can accidentally start draining another pod). Should we verify that the current amount of pods matches the expected replica slots across every pool before allowing another drain?

pkg/resource-handler/controller/shard/reconcile_pool_pods.go
```diff
@@ -346,6 +348,50 @@ func isPoolHealthy(
...
+	for i := range podList.Items {
+		pod := &podList.Items[i]
+		// DRAINED and QUARANTINED pods are expected to be unhealthy and must
+		// not block other pods from rolling.
+		if role := resolvePodRole(shard, pod.Name); role == "DRAINED" || role == "QUARANTINED" {
+			continue
+		}
+		if !pod.DeletionTimestamp.IsZero() {
+			return false, nil
+		}
```
niconosenzo: I'm thinking about the scenario where pods are stuck in `iImagePullBackOff`. This will become a deadlock state for the rolling update, not allowing to push a healthy image into the pods.
What do you think about checking for this specific error, check whether the target image is != than the current one, and allow a candidate pool to replace its own image when there is an `ImagePullBackOff` ?.

pkg/resource-handler/controller/shard/reconcile_pool_pods.go
kangmingtay: yeah this validation is currently missing from the `isPoolHealthy` function as well - we never ever check if the actual number of pods in the pool match the `replicasPerCell` count

pkg/resource-handler/controller/shard/reconcile_pool_pods.go
kangmingtay: i'm inclined to fix this separately as a stuck rollout is tangential to a rollout that results in downtime - will make a follow-up PR for this

Verolop (commented):
lgtm!

Non-blocking: the healthy pod in this test is missing the db, table-group, and shard labels used by `isShardHealthy`’s selector, so it isn’t actually included in the list. Could you pick this up in a follow-up PR? Using `buildPoolLabelsWithCell` here would keep the fixture aligned with the production selector.

Verolop (approved):

PR #626, opened 2026-09-04
fix(shard): dial multipoolers with mTLS when internalTLS is enabled

niconosenzo:
## Summary

Fix operator-to-multipooler RPCs for clusters with `spec.internalTLS.enabled: true`.

The operator now uses a cluster-specific mTLS client when calling `Status`, `GetBackups`, and `ReloadConfig`. Clusters without internal TLS continue using the existing insecure client.

## Root cause

`InternalTLS` clusters configure multipooler as an mTLS gRPC server. However, the shard controller continued using a client configured with `insecure.NewCredentials()`.

Multipooler therefore rejected every operator RPC with a connection reset. This prevented the operator from observing pooler posture and backup state or applying reload-safe configuration changes. Posture remained unknown, `PostureConsistent` became `Unknown`, and affected shards remained `Progressing`.

The posture debounce remains useful for transient observation failures, but it cannot recover from a permanently incompatible transport.

## Solution

The `MultigresCluster` certificate controller now creates a dedicated operator client certificate for every internal-TLS cluster.

The certificate:

- Is owned by the `MultigresCluster` and follows its lifecycle.
- Uses the same resolved issuer as the cluster's server certificates.
- Is issued for client authentication only.
- Is stored in the cluster namespace.

The new `poolerclient.Resolver`:

- Returns the existing insecure client when internal TLS is disabled.
- Derives the owning cluster from the shard labels.
- Reads the cluster-owned operator certificate Secret directly.
- Caches one TLS `MultipoolerClient` per cluster.
- Verifies the exact server identity: `multipooler.<cluster>.<namespace>.multigres.internal`.
- Uses standard certificate-chain, hostname, and server-auth verification.
- Re-reads the Secret periodically and rebuilds the client only when its `ResourceVersion` changes.
- Keeps the last working client if a refresh encounters a transient API failure or temporarily malformed rotated Secret.
- Retains replaced clients briefly so in-flight reconciliations can complete.
- Allows concurrent reconciliations to continue using the current client while a refresh is in progress.

This removes the need for runtime Certificate creation, issuer-specific certificate names, a separate issuer flag, and a `MultigresCluster` API lookup during every resolution.

## Failure behavior

If the client certificate has not been issued yet, the shard controller:

- Skips multipooler RPC phases.
- Publishes topology-derived pod roles.
- Sets `PostureConsistent=Unknown` with reason `PoolerClientUnavailable`.
- Keeps the shard in `Progressing`.
- Emits a `PoolerClientUnavailable` warning event.
- Requeues after 10 seconds without controller error backoff.

The existing Healthy/no-primary safety guard remains unchanged.

## Security

TLS verification is bound to the shard's exact cluster and namespace. A certificate for another cluster, even when signed by the same issuer, is rejected.

The operator client certificate is cluster-owned, so deletion or certificate reconciliation cleans it up together with the other cluster certificates.

### DIFF
--- a/api/v1alpha1/common_types.go
+++ b/api/v1alpha1/common_types.go
@@ -359,13 +359,16 @@ const (
 	// Component certificate names used for internal gRPC TLS/mTLS.
 	ComponentMultiAdminTLS   = "multiadmin"
 	ComponentMultiGatewayTLS = "multigateway"
+	ComponentMultiOrchTLS    = "multiorch"
+	ComponentMultiPoolerTLS  = "multipooler"
+	// ComponentOperatorTLS identifies the per-cluster client credential used by
+	// the operator when it calls that cluster's multipoolers.
+	ComponentOperatorTLS = "multigres-operator"
 
 	// InternalTLSIdentityDomain is the logical DNS suffix used for internal
 	// component identities. These names are verified through an explicit gRPC
 	// server-name override and are not expected to resolve through cluster DNS.
 	InternalTLSIdentityDomain = "multigres.internal"
-	ComponentMultiOrchTLS     = "multiorch"
-	ComponentMultiPoolerTLS   = "multipooler"
 )
 
 // ComponentCertSecretName returns the cert-manager Secret name for an internal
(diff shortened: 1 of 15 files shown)


Verolop (approved):
lgtm!

PR #627, opened 2026-09-04
feat(toposerver): require a client certificate on the topology server

Verolop:
## Description

this PR turns on TLS for the etcd topology server and moves every client onto it, so the `topoTLS` gate can be flipped on. etcd requires a client cert, the four pod components (multiadmin, multigateway, multipooler, multiorch) mount the issued one, and the operator loads it from the referenced Secret for its own connections. Off by default, and the whole `topoTLS` block (enablement and issuer) is immutable so it is fixed at creation: etcd can't serve plaintext and TLS on one port, its persisted peer membership isn't rewritten by a restart, and clients load their cert at process start, so neither the switch nor a CA rotation can roll out safely on a running cluster.

For multipooler the cert is the only isolation from the postgres superuser it shares a pod, IP, data PVC, and socket dir with, so a network policy won't do. It's mounted only into the topo-speaking containers, off any volume postgres or pgctld mount, with `shareProcessNamespace` unset, and a pod-spec test checks all three.

## Main changes

- etcd client and peer listeners move to https with `ETCD_CLIENT_CERT_AUTH` and `ETCD_PEER_CLIENT_CERT_AUTH`; metrics stays http for the probes. Peer TLS is included.
- Client cert mounted into the four pod components and nowhere else, with `--topo-etcd-tls-cert/-key/-ca`.
- The operator reads the credential with the uncached API reader (its cache only stores tenant Secrets carrying the managed-by label, which cert-manager and user-provided Secrets lack) and passes it as per-connection `TLSOptions`; an empty reference stays plaintext.
- `topoTLS` is immutable via a CEL rule, so enablement and issuer can only be set at creation.
- No credential issued for external topologies. Missing client material fails loudly, naming the Secret, as a Degraded phase with a `TopologyReady=False` condition that clears once the operator reconnects.

## Testing

Ran `make verify`, `make pre-commit`, and the TopoServer, cluster, shard, and topo integration suites locally, all green, including CEL integration tests that reject enabling topoTLS or changing its issuer after creation. Added the mount-boundary test, topoTLS-off render tests, and the operator secret-loading and loud-failure tests. Added a shared e2e scenario (`test/e2e/shared/topotls`) that installs cert-manager, issues from a CA so the server and clients share one root, and brings up a topoTLS cluster; a served `SELECT 1` proves the full mTLS path including the operator's own registration. Kick it with `/e2e`.

### DIFF
--- a/api/v1alpha1/multigrescluster_types.go
+++ b/api/v1alpha1/multigrescluster_types.go
@@ -50,6 +50,15 @@ import (
 // ============================================================================
 
 // MultigresClusterSpec defines the desired state of MultigresCluster.
+//
+// topoTLS is immutable after creation. Enabling it switches the etcd client and
+// peer listeners to TLS, and changing the issuer rotates the CA every certificate
+// chains to. etcd cannot serve plaintext and TLS on one port, its persisted peer
+// membership is not rewritten by a restart, and the topology clients load their
+// certificate at process start, so neither the enablement nor the issuer can be
+// changed on a running cluster without a coordinated migration. Both are fixed at
+// creation.
+// +kubebuilder:validation:XValidation:rule="has(self.topoTLS) == has(oldSelf.topoTLS) && (!has(self.topoTLS) || self.topoTLS == oldSelf.topoTLS)",message="topoTLS is immutable and can only be set when the cluster is created"
 type MultigresClusterSpec struct {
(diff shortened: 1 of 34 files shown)


rafael (approved):
Stamping based on e2e test validation.

PR #628, opened 2026-09-04
fix(shard): apply default runtime identity to any pool image

### LINKED ISSUE #466: CreateContainerConfigError on Kind: non-numeric user (postgres) with runAsNonRoot
## Current behavior

When applying MultigresCluster `minimal.yaml`, the generated pods minimal-postgres-default-0-inf-pool-default-zone-a-xxxxx-x 
- Init container multipooler succeeds.
- Main postgres container fails repeatedly: "container has runAsNonRoot and image has non-numeric user (postgres), cannot verify user is non-root".
- Pods stuck at 1/2 READY (only sidecar runs), while global-topo and multigateway works fine.

## Expected behavior

Applying `minimal.yaml` should result in all generated pool pods becoming `2/2 Running`.

## Steps to reproduce

```bash
kind create cluster
kubectl apply --server-side -f \
  https://github.com/multigres/multigres-operator/releases/latest/download/install.yaml
kubectl apply -f https://raw.githubusercontent.com/multigres/multigres-operator/main/config/samples/minimal.yaml

```

See the error:

```
NAME                                                              READY   STATUS                       RESTARTS   AGE
minimal-global-topo-0                                             1/1     Running                      0          26s
minimal-global-topo-1                                             1/1     Running                      0          26s
minimal-global-topo-2                                             1/1     Running                      0          26s
minimal-multiadmin-5d7dcff99-855bs                                1/1     Running                      0          26s
minimal-multiadmin-web-696cbd4555-zcmjn                           1/1     Running                      0          26s
minimal-postgres-default-0-inf-multiorch-zone-a-bbe2a333-68pdmt   1/1     Running                      0          15s
minimal-postgres-default-0-inf-pool-default-zone-a-a3a0d77b-0     1/2     CreateContainerConfigError   0          15s
minimal-postgres-default-0-inf-pool-default-zone-a-a3a0d77b-1     1/2     CreateContainerConfigError   0          15s
minimal-postgres-default-0-inf-pool-default-zone-a-a3a0d77b-2     1/2     CreateContainerConfigError   0          15s
minimal-zone-a-multigateway-0ccc2e8b-6bb886c8bd-nv6lf             1/1     Running                      0          16s
```

Pod events:

```
Events:
  Type     Reason     Age                  From               Message
  ----     ------     ----                 ----               -------
  Normal   Scheduled  2m47s                default-scheduler  Successfully assigned default/minimal-postgres-default-0-inf-pool-default-zone-a-a3a0d77b-0 to multigres-operator-dev-control-plane
  Normal   Pulled     2m47s                kubelet            spec.initContainers{multipooler}: Container image "ghcr.io/multigres/multigres:sha-cbc77e3" already present on machine and can be accessed by the pod
  Normal   Created    2m47s                kubelet            spec.initContainers{multipooler}: Container created
  Normal   Started    2m47s                kubelet            spec.initContainers{multipooler}: Container started
  Normal   Pulled     0s (x14 over 2m42s)  kubelet            spec.containers{postgres}: Container image "ghcr.io/multigres/pgctld:sha-cbc77e3" already present on machine and can be accessed by the pod
  Warning  Failed     0s (x14 over 2m42s)  kubelet            spec.containers{postgres}: Error: container has runAsNonRoot and image has non-numeric user (postgres), cannot verify user is non-root (pod: "minimal-postgres-default-0-inf-pool-default-zone-a-a3a0d77b-0_default(d589b246-7ba2-4b42-89f1-6328cfa3d077)", container: postgres)

```

## Immediate Workaround

Replace minimal.yaml with:

```yaml
kind: MultigresCluster
metadata:
  name: minimal
  namespace: default
spec:
  pvcDeletionPolicy:
    whenDeleted: Delete
    whenScaled: Delete
  # A minimal cluster requires at least one cell definition.
  #
  # The operator will resolve configuration defaults in this order:
  # 1. Namespace Defaults: If a template named "default" exists in this namespace
  #    (e.g., CoreTemplate/default, CellTemplate/default), it will be used.
  # 2. Hardcoded Defaults: If no template is found, the operator injects:
  #    - Default GlobalTopoServer (Managed Etcd, 3 replicas)
  #    - Default MultiAdmin (1 replica)
  #    - Default Database ("postgres") with Default TableGroup ("default")
  cells:
    - name: "zone-a"
      # zone: "us-east-1a"
  databases:
    - name: "postgres"
      default: true
      tablegroups:
        - name: "default"
          default: true
          shards:
            - name: "0-inf"
              spec:
                pools:
                  default:
                    fsGroup: 999

```

## Environment

- Host: Linux (Arch), kernel 6.18.22-1-lts, x86_64
- kind: v0.31.0
- kubectl client: v1.35.3 (kustomize v5.7.1)
- Kubernetes server: v1.35.0 (kind node image kindest/node:v1.35.0)
- Container runtime in node: containerd://2.2.0
- Docker client/server: 29.4.0 / 29.4.0
- Docker host cgroup: v2
- Operator image: ghcr.io/multigres/multigres-operator:0.10.3

graveland:
## What changed

Pool `postgres`/`multipooler` containers now always get a default numeric `RunAsUser`/`RunAsGroup` unless the CR sets one explicitly, instead of only when the image string exactly matches the pinned default.

## Why

The old gate meant the default identity applied only on an exact string match against `DefaultPostgresImage`. Any drift — an image override, or just `:main` vs. the pinned `sha-` tag — left `RunAsNonRoot` set with no `RunAsUser`. `pgctld` declares `USER postgres` by name, so Kubernetes can't verify non-root and the pod fails:

    CreateContainerConfigError: container has runAsNonRoot and image has
    non-numeric user (postgres)

Reproduces on stock `minimal.yaml`, no override needed. Closes #466. 

## Notable details / risks

`effectiveMultipoolerIdentity` had the same gate, latent only because the default multigres image happens to use numeric UID 65532 — fixed the same way, preserving the postgres/multipooler shared-UID invariant `validatePoolRuntimeIdentity` enforces. An image needing a different numeric user must now set `pool.Postgres.RunAsUser` explicitly; it's no longer inferred.

One thing I did notice because of this: our postgresql images are alpine-based, and the add-user command in them don't specify a numeric ID, but just take whatever alpine has next available. We should probably think about the user/group ID, and make it static, and audit the `postgres` user in all containers across the whole fleet and ideally consolidate them?


Verolop (commented):
Thanks, this makes sense! 
Before merging, can you please update the API comments and generated CRDs too? They still say unset IDs use the image defaults. While you're there, also mention that custom images may now need explicit runAsUser and runAsGroup, especially with existing volumes.

regarding auditing IDs, yes please. 
Also note Alpine is only the builder, and the Debian img already sets postgres to 999:999, so maybe keep that for the containers sharing PGDATA and handle the broader audit separately, WDYT?

graveland:
> Thanks, this makes sense! Before merging, can you please update the API comments and generated CRDs too? They still say unset IDs use the image defaults. While you're there, also mention that custom images may now need explicit runAsUser and runAsGroup, especially with existing volumes.

Good catch! I pushed a doc update.

> regarding auditing IDs, yes please. Also note Alpine is only the builder, and the Debian img already sets postgres to 999:999, so maybe keep that for the containers sharing PGDATA and handle the broader audit separately, WDYT?

yeah, auditing is definitely separate!

Verolop (approved):
lgtm!

PR #629, opened 2026-09-07
test(shard): avoid integration update conflicts

niconosenzo:
Patch inline configuration changes without replacing the complete Shard object, which can race with concurrent controller status updates.


Verolop (approved):
lgtm!

PR #630, opened 2026-09-07
feat(shard): let multipooler derive connpool capacity from max_connections

haritabh17:
## Summary

Upstream multipooler ([multigres/multigres#1439](https://github.com/multigres/multigres/pull/1439)) now derives its global connection pool capacity from the server's `max_connections` at admin-pool open when `--connpool-global-capacity` is not set, and re-derives it after every postgres restart.

- Remove the hardcoded `--connpool-global-capacity=40` / `--connpool-admin-capacity=5` flags added in #474. The admin capacity of 5 was already the upstream default. The pool now tracks whatever `max_connections` the operator renders, including user overrides in the Shard spec.
- Bump default images and the `multigres` module to `sha-a8e8aab`, the current upstream main head and the first pin that includes the derivation. The previous pin (`sha-b0d11a0`) predates it, so the flag removal alone would have regressed to the upstream default of 100 against a postgres with 60.

Closes MUL-1527.

## Test plan

- [x] `go build ./...`, `go test ./pkg/resource-handler/controller/shard/ ./api/...`
- [ ] kind e2e: confirm multipooler logs `derived connpool global capacity from postgres` with the new image


niconosenzo (approved):
lgtm.

PR #632, opened 2026-09-09
fix(postgresconfig): classify backend/superuser-backend GUCs as restart

GuptaManan100:
## Summary

- Backend and superuser-backend context GUCs (`log_connections`, `log_disconnections`, and four others) were classified reload-safe, but a SIGHUP reload never applies them to the multipooler's long-lived pooled backends — so the change silently never took effect, while the pooler's `ReloadConfig` still reported success.
- Classify `backend`/`superuser-backend` as restart so these GUCs go through pod recreation, which recycles the pooled backends and actually applies the change.
- Add an e2e regression test that enables `log_connections` on a running cluster and asserts it takes effect through the gateway.

## Problem

`log_connections` set via `spec.postgresConfig` had no effect on a running cluster. It is a `superuser-backend` context GUC: its value is fixed when a backend starts, and a SIGHUP reload only changes the value the postmaster hands to subsequently-started backends, never existing ones (see `set_config_option`'s `PGC_BACKEND` handling in PostgreSQL). Because the multipooler fronts PostgreSQL with long-lived pooled backends, no new backend is started on a reload, so the change never reaches a running backend.

The failure was silent: the pooler's `ReloadConfig` gate checks `pg_file_settings.applied`, which is `true` for these GUCs, and only escalates `needs_restart` for `postmaster` context — so the operator marked the config change done while it had no effect.

## Solution

Add `backend` and `superuser-backend` to the classifier's `restartContexts`, so these GUCs drive pod recreation (the restart path) instead of an in-place reload. Recreating the pods restarts PostgreSQL with the new config, which is the only way the value reaches the pooled backends. Six GUCs are affected: `log_connections`, `log_disconnections`, `jit_debugging_support`, `jit_profiling_support`, `ignore_system_indexes`, `post_auth_delay`.

Verified end-to-end on Kind (reproduce-then-fix): with the unfixed operator the new `TestLogConnectionsTakesEffect` fails (`SHOW log_connections` stuck at `off`); with the fix it passes (pods recreated, value becomes `on`). `sighup`/`user`/`superuser` GUCs (e.g. `work_mem`, `max_wal_size`) remain reload-safe and unaffected.

### DIFF
--- a/test/e2e/shared/postgresconfig/logconnections_test.go
+++ b/test/e2e/shared/postgresconfig/logconnections_test.go
@@ -6,10 +6,6 @@ import (
 	"context"
 	"testing"
 
-	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/types"
-	"sigs.k8s.io/controller-runtime/pkg/client"
-
 	"github.com/multigres/multigres-operator/test/e2e/framework"
 )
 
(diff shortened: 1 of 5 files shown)


test/e2e/shared/postgresconfig/logconnections_test.go
```diff
@@ -0,0 +1,109 @@
...
+		if err := c.List(ctx, pods, client.InNamespace(ns),
+			client.MatchingLabels{"app.kubernetes.io/component": "shard-pool"}); err != nil {
+			t.Fatalf("list pool pods: %v", err)
+		}
+		uids := map[string]types.UID{}
+		for i := range pods.Items {
+			uids[pods.Items[i].Name] = pods.Items[i].UID
+		}
+		return uids
+	}
```
niconosenzo: nit: seems we define this same closure 3 times on the same package, wdyt about pulling it to package level func ? e.g.
```golang
func poolPodUIDs(t *testing.T, c client.Client, ctx context.Context, ns string) map[string]types.UID {
###
}
```

niconosenzo (approved):
left a nit comment, but it lgtm!

test/e2e/shared/postgresconfig/logconnections_test.go
GuptaManan100: Fxied

PR #635, opened 2026-09-10
fix(postgresconfig): render baseline after ref so operator sizing wins

GuptaManan100:
## Summary

- `Render()` appended config layers `baseline → ref → inline`, so the deprecated `PostgresConfigRef` overrode the operator's resource-derived baseline (PostgreSQL applies later assignments last-write-wins).
- This is safe for a ref that's a few direct GUC lines, but a ref is **opaque raw text**: if it contains an `include` directive, PostgreSQL transitively pulls in an external file the operator never sees, silently overriding carefully sized values (`shared_buffers`, `effective_cache_size`, `max_slot_wal_keep_size`, …) with whatever is baked into that file.
- Swap to `ref → baseline → inline` so the operator's own sizing math always wins over the deprecated ref, while inline `spec.postgresConfig` remains the single explicit override.

## Problem

The operator treats `refContent` as opaque text appended after its baseline "so the user's override wins." But when that text is e.g. `include = '/etc/postgresql/postgresql.conf'`, Postgres expands it at parse time into a whole external file the operator has zero visibility into — and that file can redefine the operator's resource-derived GUCs, silently replacing them with generic image defaults. Symptom: the operator's sizing looks correct in isolation and in inline-only tests, yet a real project's baseline values never take effect.

## Solution

Render the baseline **after** the ref: `ref → baseline → inline`. The operator's resource-derived values (and anything the baseline sets) now win over the ref and anything it transitively includes; inline `spec.postgresConfig` stays last and remains the one intended way to deviate from the baseline.

This is a deliberate **precedence semantics change** for the deprecated ref (baseline-vs-ref inverts). `ref`-vs-`inline` is unchanged — inline still wins.

### DIFF
--- a/pkg/postgresconfig/render.go
+++ b/pkg/postgresconfig/render.go
@@ -1,9 +1,13 @@
 // Package postgresconfig renders the effective postgresql.conf the operator
-// mounts into pgctld. The operator owns config generation end-to-end: a static
-// baseline it defines, followed by the user's legacy PostgresConfigRef content
-// and the inline spec.postgresConfig map, each appended so it overrides earlier
-// layers (PostgreSQL applies later assignments last-write-wins). Resource-
-// derived sizing is baked into the Config before rendering.
+// mounts into pgctld. The operator owns config generation end-to-end. Layers are
+// appended in order of increasing precedence (PostgreSQL applies later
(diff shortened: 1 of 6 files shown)


niconosenzo (approved):
lgtm!

GuptaManan100:
Verified E2E in staging!

PR #643, opened 2026-09-11
feat(operator): derive GOMEMLIMIT and expose pprof on loopback

graveland:
## What changed

- pprof is on in the shipped manifest: `config/manager/manager.yaml` passes
  `--pprof-bind-address=127.0.0.1:6060`. The flag itself defaults to empty, so
  a bare `./manager` stays off and does not claim the port during local runs.
- `GOMEMLIMIT` derived at startup from `MEMORY_LIMIT_BYTES`, projected from
  `limits.memory` via the downward API, at 90% of the limit.
- Memory limit `512Mi`, requests `100m`/`256Mi`, no CPU limit. These replace
  untouched kubebuilder scaffold values (`128Mi`/`500m`), not measured ones.

## Why

Go derives `GOMAXPROCS` from the cgroup CPU quota but has no equivalent for
memory (golang/go#75164), so without `GOMEMLIMIT` the collector has no reason
to run before the cgroup OOM-kills the process.

pprof is on by default because it cannot usefully be enabled on demand:
changing args rolls the Deployment, so the process holding the heap under
investigation is replaced by a fresh one. A slow leak takes days to reappear,
and after an OOMKill the evidence is already gone.

## Notable details / risks

- **Access control is delegated to Kubernetes on purpose.** Binding to loopback
  means the only route in is the API server's `pods/portforward` subresource, so
  reaching the endpoint requires a principal that authenticates to the cluster
  and that RBAC authorizes for portforward in this namespace, and the attempt is
  recorded in the API server audit log. The endpoint stays unauthenticated
  because adding a check there would duplicate one that already happened, while
  introducing a second credential to issue, rotate and leak. Revocation stays
  where cluster access is already managed.

  What is actually exposed, to be precise about it: profiles are sampled stack
  traces and byte counts, so they reveal code paths and allocation sizes, not
  the contents of memory. The endpoint does serve this process's argv via
  `/debug/pprof/cmdline`, and `/debug/pprof/profile` and `/trace` let any caller
  that reaches the port consume CPU on demand, which matters for a
  leader-elected singleton.
- No CPU limit is deliberate: five controllers run at `MaxConcurrentReconciles`
  20, and throttling the collector is the wrong way to fail when `GOMEMLIMIT` is
  what keeps the process inside its memory limit. A CPU limit also sets
  `GOMAXPROCS` (500m pinned it to 2).
- The `512Mi` floor is arithmetic, not empirical: `GOMEMLIMIT` excludes the
  binary's own mapping (~58MiB stripped), so at `128Mi` the 90% handed to the
  runtime could not fit alongside it. Resource values are provisional pending
  the profiles this PR makes possible.
- `.gitignore`: the bare `multigres-operator` entry matched any path component
  and was hiding `cmd/multigres-operator/`, which blocked `main_test.go`.
  Anchored to `/multigres-operator` rather than deleted, since it still guards
  against an ad-hoc `go build ./cmd/multigres-operator` dropping a binary at
  the repo root.


cmd/multigres-operator/main.go
```diff
@@ -193,6 +206,33 @@ func main() {
...
+	// (go1.25+), but has no equivalent for memory: golang/go#75164 is still an
+	// open proposal, so GOMEMLIMIT defaults to math.MaxInt64 and a reconcile
+	// burst OOMKills instead of making the collector work harder. Derive it
+	// from the limit the downward API projects in manager.yaml.
+	rawMemLimit := os.Getenv("MEMORY_LIMIT_BYTES")
+	// An absent value is the ordinary case outside a container and needs no
+	// comment: "resolved runtime limits" below reports the unlimited default.
+	// A malformed one means someone tried to set a limit and silently did not.
+	if memLimit, ok := goMemLimitFromEnv(rawMemLimit); ok {
+		debug.SetMemoryLimit(memLimit)
```
Verolop, line 219: Can we keep GOMEMLIMIT if someone has already set it? They might have chosen a lower value to leave more headroom, and we’d overwrite it here since the manifest always supplies MEMORY_LIMIT_BYTES.

Verolop (commented):
nice! I just have a small ux request.

Verolop (approved):

PR #653, opened 2026-09-14
fix(deps): go mod tidy observer tool

niconosenzo:


Verolop (approved):
I always forget about the observer...  :D

PR #656, opened 2026-09-16
fix(shard): recheck cohort before PVC deletion

niconosenzo:
## Description
Before hard-deleting an excess PVC during scale-down, re-read topology and pooler consensus state to confirm the removed pooler is absent from the committed cohort. The pooler's identity is rebuilt from its `--service-id` so it matches real CohortMembers entries. While absence cannot be proven (or the pooler is still a member) the pod and PVC are retained and the reconcile requeues, the first deferral is stamped on the pod. If confirmation still fails after 10 minutes the PVC is orphaned for retention-window GC instead of deleted and the pod is released (Warning PVCCleanupFallback), so a persistent failure cannot wedge scale-down. DRAINED (data known-bad) pods skip the recheck: orch already ejected them and holding the pod would pin the ordinal slot. Only the topology primary is queried.

## Changes
- Add a final cohort-membership check immediately before permanently deleting an excess PVC.
- Query live topology and pooler status to establish the committed cohort and confirm the target pooler is absent.
- Fail closed when topology, RPC, identity, primary, or consensus evidence is missing or conflicting.
- Defer PVC deletion when safety cannot be confirmed.
- Preserve the ready-for-deletion pod so a later reconciliation can retry cleanup.
- Mark the rollout as waiting for recovery, triggering the existing five-second requeue.
- Emit a PVCDeletionDeferred event explaining why cleanup was postponed.
- Keep the existing orphan-retention path unchanged, the new check only protects immediate hard deletion.
- Add tests covering successful deletion, cohort-member deferral, unavailable or invalid authority, and conflicting primary observations.
- New pod annotation `drain.multigres.com/pvc-cleanup-deferred-since`.
- Only the topology primary is queried now (dead replicas no longer burn the 30s budget).
- New Warning event reason PVCCleanupFallback.

## Testing
```bash
go test ./pkg/data-handler/posture ./pkg/resource-handler/controller/shard
```


niconosenzo:
Closing for now. We will take a different approach regarding PVC deletion.

PR #662, opened 2026-09-18
feat(multigrescluster): track sticky InitializedAt on first Healthy phase

niconosenzo:
## Description
Adds status.InitializedAt, set once when a cluster first reaches PhaseHealthy and never cleared on later Degraded transitions. Exposes it as an `initialized` label on multigres_operator_cluster_info so alerting can distinguish first-time bootstrap from a real incident on an established cluster.

## Summary of changes
- `MultigresClusterStatus.InitializedAt *metav1.Time` field, set once, sticky (first time phase reaches Healthy, never cleared).
- `status.go`: sets it right after phase computation, before the server-side patch.
- `multigres_operator_cluster_info` gauge: new initialized label ("true"|"false"), SetClusterInfo signature updated, controller call site updated.
- `pkg/monitoring/doc.go` example + tests updated for the new signature.
- New test `TestUpdateStatus_InitializedAtSticky` proving it survives a later Degraded transition.


pkg/cluster-handler/controller/multigrescluster/status.go
```diff
@@ -252,6 +252,12 @@ func (r *MultigresClusterReconciler) updateStatus(
 		cluster.Status.Message = "Cluster is progressing"
 	}
 
+	if cluster.Status.Phase == multigresv1alpha1.PhaseHealthy &&
```
Verolop, line 255: Can you please check that the expected children exist and are healthy before setting this? The phase calculation treats empty child lists as Healthy, so if the cache hasn’t caught up yet, a cluster initialized before anything is ready can be marked as Healthy

niconosenzo: Good catch, Should we change that and treat them as "progressing" ?  The way I see it, if "Healthy" child doesn't mean really healthy, we will have to add safeguards on other places where we depend on that phase, correct?

niconosenzo: > if the cache hasn’t caught up yet, a cluster initialized before anything is ready can be marked as Healthy

Actually, not sure if I understood this part, the flag goal is only to avoid triggering alerts on clusters that are not initialized, but has little to do with the healthy status of the cluster.

Verolop: yeah, agreed! Let’s keep it Progressing until all expected children are present and healthy. Let's also check for missing children too.

Verolop (approved):

PR #663, opened 2026-09-18
fix(shard): always orphan PVCs on scale-down instead of hard-deleting

niconosenzo:
## Description

Removes the replica-count threshold that decided between orphaning a PVC (deferred deletion via multigres-gc) and hard-deleting it in-line during pool scale-down and shard removal. 

Now: PVC cleanup during scale-down/drain/replace/shard-removal always orphans, unconditionally. The only case that still hard-deletes in-line is when the owning MultigresCluster itself is being deleted, at that point there's nothing to roll a scale-down back to, so there's no reason to wait out the retention window.

## Changes

- Remove `pvcOrphanReplicasThreshold` / `orphanByRemainingCount`  (the hard-delete-past-threshold logic) from shard_controller.go.
- Add clusterIsChurning(): looks up the owning MultigresCluster and reports whether it's being deleted. If the cluster is missing entirely, fails safe and treats it as not churning (orphan), since it should still exist with its cleanup finalizer for as long as its Shards are being torn down.
- cleanupShardPVCs (Shard deletion path): always orphans, except when clusterIsChurning,  then deletes in-line. Drops the pool+cell grouping/sorting that only existed to support the threshold decision.
- cleanupPodPVC (pool pod scale-down/drain/replace path): always orphans now, this path never runs during cluster teardown (Shard reconcile short-circuits to handleDeletion once the Shard has a deletion timestamp), so no churn check is needed here.


pkg/resource-handler/controller/shard/shard_controller.go
```diff
@@ -39,24 +39,29 @@ const (
...
+func (r *ShardReconciler) clusterIsChurning(
+	ctx context.Context,
+	namespace, clusterName string,
+) (bool, error) {
+	if clusterName == "" {
+		return false, nil
+	}
+	cluster := &multigresv1alpha1.MultigresCluster{}
+	err := r.Get(ctx, client.ObjectKey{Namespace: namespace, Name: clusterName}, cluster)
+	if errors.IsNotFound(err) {
```
Verolop, line 58: I think the cluster controller drops its finalizer before the shards finish cleaning up. If the cluster is already gone by the time we get here, we’ll get NotFound and orphan the PVCs anyway. We can orphan here too unless you want to make the cluster wait for shard cleanup.
I tested this by running the cluster deletion reconciliation before shard cleanup. The parent disappeared, and the PVC remained with an orphan timestamp and no owner references.

Verolop (approved):

